PR #4, opened 2021-05-28
Sinks: Add logger generator

jli35:
Add logger.py to enable log call to create working logger

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/4)

### DIFF
--- a/sinks/logs/logger.py
+++ b/sinks/logs/logger.py
@@ -1,0 +1,38 @@
+# Generates loggers based on channel passed to log
+import logging
+
+def log(channel, fname = 'globalLog.log'):
+    """
+    Return a new logger
+
+    Parameters
+    ----------
+
+    channel: str
+        The channel the message on the bus is being passed through.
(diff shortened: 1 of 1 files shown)


BluCodeGH (changes requested):
Sorry, it looks like the requirements for this weren't explained very well, let me try to clarify.

Each source and sink is an independent python program, and they communicate with each other using the omnibus library. The logger should be a sink (ie an independent python program) that subscribes to all messages (using the omnibus library) and writes them to a file.

You will want to instantiate a `Receiver` listening on the channel `""` (all messages) and in a loop call `recv_message` (see `omnibus/omnibus.py` docstrings). Then using the timestamp, channel and payload data in the message write a line to the log file. For now it is fine to use `str(payload)` since we want the logs to be human-readable during testing, but in the future we will want to change them to be machine readable so that they can be replayed.

Issue #32, opened 2021-09-23
Log Replay


BluCodeGH:
Add the ability to play back past logs in real-time. Even better would be the ability to change playback speed, pause/resume and jump around.

BluCodeGH:
Migrated to https://app.clickup.com/t/1kqp4kf

PR #26, opened 2021-07-31
Omnibus: Add unit testing

BluCodeGH:
Closes #15 and speeds up server IP discovery.

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/26)

PR #34, opened 2021-10-16
ReplayLog: Initial commit for Replay Log source (#26)

DevinLeamy:
Implemented a Replay Log source as described in https://github.com/waterloo-rocketry/omnibus/issues/32.

**Features:**
- Real time log replay.
- `(-r, --replay_speed)` flag (_default: 1x_)
- `(-m, --max_logs)` flag (_maximum number of log files to display when prompted_)
- `(-l, --log_file)` flag (_the relative path to a log file (default: prompt the user with the global log sink's log files as options_)

To try it out, feed some data into the Global Log sink `globallog/main.py` to create a log file, and then run `replay_log/main.py`.

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)


BluCodeGH (changes requested):
A solid start, and great to see you know your python. One thing that's notably missing are unit tests, you'll probably want to mock out the sender using pytest. I do have some general thoughts on the architecture:

I'm not sure a class is the way to go here. Classes are primarily useful as a way to tie functions to the data they work on, especially when you will have lots of different instances with different data. In this case I don't really see a use case for more than one instance, and perhaps more damning is that the extent of the data is a filename and a speed which are represented perfectly well as function arguments.

I would suggest instead exposing a handful of functions (really it boils down to `replay_selected_log`) that form the library's API with their arguments.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-:-Mm6RF84E5JEL8ykI8fE:b-cyl1d0)* status: all files reviewed, 10 unresolved discussions (waiting on @DevinLeamy)

___
*[sinks/globallog/main.py, line 17 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6J3BI2DVTCY4X0TQQ:-Mm6J3BI2DVTCY4X0TQR:b-tpgho1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sinks/globallog/main.py#L17)):*
> ```Python
> receiver = Receiver(CHANNEL)
> # Creates new file
> with open(fname, "ab") as f:
> ```

Given that we use the date and time for the filename I'd really hope we never need to append? A check for if the file exists would be a better alternative, but I'm not sure it's strictly necessary.
___
*[sinks/globallog/main.py, line 21 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6JD-d1uDB_b51NWd4:-Mm6JD-d1uDB_b51NWd5:bux79dc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sinks/globallog/main.py#L21)):*
> ```Python
>         msg = receiver.recv_message()
>         f.write(msgpack.packb([msg.channel, msg.timestamp, msg.payload]))
>     f.close()
> ```

`with` blocks close the file automatically, like a try finally would (ie even if there is an exception).
___
*[sources/replay_log/main.py, line 1 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6JStz0KBB1BJ_wufn:-Mm6JStz0KBB1BJ_wufo:b-yeo7p8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/main.py#L1)):*
> ```Python
> from sources.replay_log.replay_log import ReplayLog
> ```

No need to specify the full path (which also assumes the directory the user is running this from), better to use the local `from replay_log import ReplayLog`.
___
*[sources/replay_log/replay_log.py, line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6K7ZB2P06cpBUux7z:-Mm6K7ZB2P06cpBUux8-:bse1oka) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L15)):*
> ```Python
> import msgpack
> from datetime import datetime
> import argparse
> ```

Take a look at the import order section of the pep8 style guide https://www.python.org/dev/peps/pep-0008/#imports. The global log sink is a good example of the format.
___
*[sources/replay_log/replay_log.py, line 23 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6KWCN3hCjmw-byp5-:-Mm6KWCN3hCjmw-byp50:b-c8p7d2) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L23)):*
> ```Python
>     args = self.parse_arguments()
>     self.MAX_LOGS = args.max_logs
>     self.REPLAY_SPEED = max(1, args.replay_speed)
> ```

Why not let us play back logs slower than 1x?
___
*[sources/replay_log/replay_log.py, line 48 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6LeOp0kPwS2I4jFNZ:-Mm6LeOp0kPwS2I4jFN_:beakn5y) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L48)):*
> ```Python
>       # wait for the real time to catch up to the delta time
>       wait_time = max(0, delta_l_time - delta_r_time)
>       time.sleep(wait_time)
> ```

Unfortunately `time.sleep` doesn't have sufficient resolution (operating systems impose a minimum sleep duration) for our purposes, instead the only way to do it is to busy loop until the times are correct. On the bright side that does let you simplify this code a lot since you don't need to worry about deltas.
___
*[sources/replay_log/replay_log.py, line 56 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6N6oF6h9_GzdJrRid:-Mm6N6oF6h9_GzdJrRie:b-m7uaxu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L56)):*
> ```Python
>       sender.send_message(message)
> 
>   def parse_arguments(self):
> ```

Generally argument parsing should happen in `main.py`. Libraries should expose simple APIs that correspond to those arguments while maintaining reusability elsewhere (ie could be called by other code without explicit command line arguments).
___
*[sources/replay_log/replay_log.py, line 70 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6NgqaANecFO1uzJyo:-Mm6NgqaANecFO1uzJyp:b-59hvg8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L70)):*
> ```Python
>     return args
> 
>   def get_replay_log(self, max_logs):
> ```

As much as I hate to say this, and it's entirely my fault for not being clearer on the requirements, this very nice code is likely unnecessary. The general use of log replay is going to be replaying a log that was downloaded into some unknown folder, and so the most ergonomic use is going to be passing that file as a mandatory command line argument.
___
*[sources/replay_log/replay_log.py, line 99 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6Oz_N3G03by4Kethe:-Mm6Oz_N3G03by4Kethf:b7wgjew) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L99)):*
> ```Python
>     return os.path.join(self.GLOBAL_LOGS, log_files[selection])
> 
>   def unpack_log(self, log_file):
> ```

Our log files after a test will be on the order of 10 Gb. We don't want to put all that in RAM. You'll instead need to directly iterate over the unpacker in `replay_selected_log`, as in `for msg in msgpack.Unpacker(f)`. Iterating directly also means you don't need that ugly out of data check, which is nice.
___
*[sources/replay_log/replay_log.py, line 120 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mm6Qyt-99kDFDaUP6fQ:-Mm6Qyt-99kDFDaUP6fR:b-a11dj6) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/24a856f26c6c646913bf606fe4f30e87be345ced/sources/replay_log/replay_log.py#L120)):*
> ```Python
>   
>   def get_global_log_files(self):
>     return [file for file in os.listdir(self.GLOBAL_LOGS) if file.split(".")[-1] == "log"]
> ```

Not an issue but you should check out the `pathlib` library, it's built in and way nicer to use than `os`!

DevinLeamy:
`pytest-mock` is one of the coolest APIs I've come across in awhile (and it's extremely useful for testing)! 

Couple of things worth noting:
- I added some tests
- and, I added `flake8` to the`./tools/format.sh` (if you're preparing a commit, mine as well have the auto-formatter check for the same things as the Action; saves a potential failed check)

BluCodeGH (changes requested):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)* status: all files reviewed, 22 unresolved discussions (waiting on @DevinLeamy)

___
*[requirements.txt, line 4 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVtDom9vepIMASaLI4:-MmVtDom9vepIMASaLI5:b-arjf80) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/requirements.txt#L4)):*
> ```help
> msgpack==1.0.2
> pytest==6.2.4
> pytest-mock==3.6.1
> ```

Although the pytest-mock API is pretty nice we can get the same functionality out of the built in `monkeypatch` fixture which saves a dependency and is consistent with the other tests.
___
*[sources/replay_log/main.py, line 6 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVgvYM4If3QhmncjJh:-MmVgvYNA-Ce-ikxQQPG:bepctdm) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L6)):*
> ```Python
> import os
> 
> import replay_log as ReplayLogSource
> ```

Nit, modules should be left as `snake_case`, `UpperCamelCase` is reserved for classes.
___
*[sources/replay_log/main.py, line 17 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVdCNg7J7pcVqtCmBi:-MmVdCNh9jOqGuy4tl0P:b-jbd9c7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L17)):*
> ```
>     parser.add_argument('--log_file', '-l', default=None,
>                         help="relative path to a log file (default: selection from prompt)")
> ```
Would be cleaner to make this an optional argument without a flag, after the other flags.
___
*[sources/replay_log/main.py, line 23 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVdTRhBV7fWwO9R0U2:-MmVdTRhBV7fWwO9R0U3:b-ww0hge) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L23)):*
> ```
>     args = parser.parse_args()
>     return args
> ```
Nit, can be combined into one line
___
*[sources/replay_log/main.py, line 37 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVeFtD4JCfYZH3LrjZ:-MmVeFtD4JCfYZH3Lrj_:b-g01ja) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L37)):*
> ```Python
> 
>     if len(log_files) == 0:
>         return None  # no global log files to replay
> ```

You `return None` here but don't have a catch for it in the main code, if there are no logs it should print an error message and exit.
___
*[sources/replay_log/main.py, line 53 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVetET5l5WJ4kEdNSr:-MmVetET5l5WJ4kEdNSs:b7sik5b) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L53)):*
> ```Python
>         except:
>             print("Error: invalid selection. Showing most recent log.")
>             selection = 0
> ```

This should be a while loop so that the user can retry if they entered an invalid number.
___
*[sources/replay_log/main.py, line 55 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVejwN0IJtylWEUnhX:-MmVejwN0IJtylWEUnhY:b6bhqna) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L55)):*
> ```Python
>             selection = 0
> 
>     return os.path.join(GLOBAL_LOGS, log_files[selection])
> ```

Nit, `Path(GLOBAL_LOGS) / log_files[selection]` (also maybe just declare GLOBAL_LOGS as a Path object.)
___
*[sources/replay_log/main.py, line 60 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVfoc-ELhUygTo4Tga:-MmVfoc-ELhUygTo4Tgb:b-swab26) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L60)):*
> ```Python
> def get_datetime(filename):
>     """
>     Assumes file was produced by globallog/main.py.
> ```

Not a safe assumption, the program shouldn't crash if it encounters a different filename (eg I like to rename log files to keep track of them). Investigate instead looking at the 'last modified time' of the file stored by the operating system.
___
*[sources/replay_log/main.py, line 70 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVg6VG7AhidbZgiuO_:-MmVg6VG7AhidbZgiuOa:b-c5sxtd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L70)):*
> ```Python
> if __name__ == "__main__":
>     args = parse_arguments()
>     MAX_LOGS = args.max_logs
> ```

Nit, I've generally seen ALL_CAPS in python be similar to #defines in C, compile-time constants. These are only runtime constants so should be lower_case.
___
*[sources/replay_log/main.py, line 71 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVgNFA9bq48B_cwddS:-MmVgNFA9bq48B_cwddT:boeix18) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L71)):*
> ```Python
>     args = parse_arguments()
>     MAX_LOGS = args.max_logs
>     REPLAY_SPEED = args.replay_speed if args.replay_speed > 0 else 1
> ```

We should tell the user explicitly via an error message that we don't support negative / zero speeds rather than silently ignoring them.
___
*[sources/replay_log/replay_log.py, line 21 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVi5Yg4zntXpyCOfXr:-MmVi5Yg4zntXpyCOfXs:b-a35yz1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/replay_log.py#L21)):*
> ```Python
>     Replays the contents of a log_file
>     """
>     unpacker = None
> ```

Nit but you set unpacker in two lines, unnecessary.
___
*[sources/replay_log/replay_log.py, line 32 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVil847lJ3Cva4jxTi:-MmVil847lJ3Cva4jxTj:b7l5alr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/replay_log.py#L32)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>         while True:
>             message = fetch_message(unpacker)
>             if message == None:
>                 break
> ```
</details>

`for channel, timestamp, payload in unpacker:`
___
*[sources/replay_log/replay_log.py, line 58 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmViy5K9bN25Xg3-2BY:-MmViy5K9bN25Xg3-2BZ:b-bu2akb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/replay_log.py#L58)):*
> ```Python
> 
> def wait_for_logtime(message, r_start_time, l_start_time, replay_speed):
>     r_delta = (time.time() - r_start_time) * replay_speed
> ```

Nit, avoid duplicated code by initializing r_delta to 0.
___
*[sources/replay_log/test_replay_log.py, line 13 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVqqog3TJgZg8u_xTP:-MmVqqog3TJgZg8u_xTQ:b-scwbvc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L13)):*
> ```
> TEST_LOG_IN = "./sources/replay_log/test_log_in.log"
> TEST_LOG_OUT = "./sources/replay_log/test_log_out.log"
> ```
Doing any more file I/O from unit tests than reading test data feels weird, the filesystem is very 'messy' in that there a ton of ways for things to go wrong that unit tests ideally avoid. I think a good move here would be to (in `replay_log.py`) abstract the opening of the file into `main.py` (take a file-like object) so that your tests can pass a `BytesIO` instead.
___
*[sources/replay_log/test_replay_log.py, line 18 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVuStG8Bqo9fSrPwYh:-MmVuStG8Bqo9fSrPwYi:beb04yw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L18)):*
> ```Python
> class TestReplayLog:
>     @pytest.fixture
>     def sender(self, mocker):
> ```

You can define the mocked function inside the sender function and have it mutate an object that sender returns, that way the test functions can get the sent messages by looking at the value of sender.
___
*[sources/replay_log/test_replay_log.py, line 58 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVrtX9-Xyeer9wTI9Z:-MmVrtX9-Xyeer9wTI9_:bclx4n5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L58)):*
> ```Python
>         t2 = time.time() - s2
> 
>         # timing is not very percise, 4x speed should be at least 2x faster
> ```

Nit "precise", also that seems like a pretty bad timing difference - I think it might be due to your `mock_recieve_message`doing file I/O.
___
*[sources/replay_log/test_replay_log.py, line 70 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVs7oHD80MsbZ1ElZx:-MmVs7oHD80MsbZ1ElZy:b-d8gbje) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L70)):*
> ```Python
>         t2 = time.time() - s2
> 
>         # timing is not very percise, a quarter to speed should be at least half as fast
> ```

Precise
___
*[sources/replay_log/test_replay_log.py, line 81 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVsDD6CSLRHbBfzBgz:-MmVsDD6CSLRHbBfzBh-:b6weitc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L81)):*
> ```Python
> 
> 
> def mock_receive_message(msg):
> ```

You never mock out receiving a message so this can be combined with the mock_send_message.
___
*[sources/replay_log/test_replay_log.py, line 84 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVsHr715CMcgaZaYYd:-MmVsHr715CMcgaZaYYe:b-b9g5x9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L84)):*
> ```Python
>     if msg == None:
>         return
>     with open(TEST_LOG_OUT, "ab") as f:
> ```

This should go away with the switch to `BytesIO` but it's very bad practice to open and close the file every time you want to write to it, especially when you're writing to it perhaps every 0.01 seconds.
___
*[sources/replay_log/test_replay_log.py, line 88 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVspCx8FjJZ-v1_YSy:-MmVspCx8FjJZ-v1_YSz:b827qmn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L88)):*
> ```Python
> 
> 
> def logs_match(in_log, out_log):
> ```

You should be able to do a byte-wise comparison rather than unpacking and checking the messages, right?
___
*[sources/replay_log/test_replay_log.py, line 90 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVseHg1fkWZ-9fhIH-:-MmVseHg1fkWZ-9fhIH0:b7f77sr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L90)):*
> ```
>     with open(in_log, "rb") as f1:
>         with open(out_log, "rb") as f2:
> ```
Nit, you can put two open statements in one with statement, separate them with a comma.
___
*[tools/format.sh, line 22 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVjeMMEhUDlndhVGwZ:-MmVjeMMEhUDlndhVGw_:b-2efk1f) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/tools/format.sh#L22)):*
> ```
> # Copied from CL
> # ---
> # stop the build if there are Python syntax errors or undefined names
> ```
Nit, You don't really need the "copied from" and "---" lines, also "stop the build" isn't correct in this context.

DevinLeamy (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @BluCodeGH and @DevinLeamy)

___
*[requirements.txt, line 4 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVtDom9vepIMASaLI4:-MnbojQy2RIZqsnboDKL:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/requirements.txt#L4)):*

Although the pytest-mock API is pretty nice we can get the same functionality out of the built in `monkeypatch` fixture which saves a dependency and is consistent with the other tests.
</blockquote></details>

Done.
___
*[sources/replay_log/test_replay_log.py, line 58 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVrtX9-Xyeer9wTI9Z:-MnbtZ9ZCvzK6wsNregj:ba97cb3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L58)):*
<details><summary><i>Previously, DevinLeamy (Devin Leamy) wrote…</i></summary><blockquote>

The timing isn't actually all that bad - I just decided to set a big margin for error. The new margin is 0.1
</blockquote></details>

Edit: To pass tests, the actual speed must be within 2% of the expected speed.

DevinLeamy (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @BluCodeGH and @DevinLeamy)

___
*[sources/replay_log/test_replay_log.py, line 18 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVuStG8Bqo9fSrPwYh:-Mnc9q1z8VxmH5lktRNU:b975w8j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L18)):*

You can define the mocked function inside the sender function and have it mutate an object that sender returns, that way the test functions can get the sent messages by looking at the value of sender.
</blockquote></details>

The current approach seems pretty idiomatic from what I've seen.
___
*[sources/replay_log/test_replay_log.py, line 81 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVsDD6CSLRHbBfzBgz:-MncA9xF5AvQShyI9Afn:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L81)):*

You never mock out receiving a message so this can be combined with the mock_send_message.
</blockquote></details>

Done.
___
*[sources/replay_log/test_replay_log.py, line 84 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVsHr715CMcgaZaYYd:-MncB59U-LocFqgn2Wda:b68mxn4) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L84)):*

This should go away with the switch to `BytesIO` but it's very bad practice to open and close the file every time you want to write to it, especially when you're writing to it perhaps every 0.01 seconds.
</blockquote></details>

Given that the files are small (because they are test files and they should be small) this should be a Nit. 

And, as you said, if the switch to a BytesIO is made this will disappear straight away.

DevinLeamy (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @BluCodeGH)

___
*[sources/replay_log/test_replay_log.py, line 13 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVqqog3TJgZg8u_xTP:-MncChqM35NiQ7lPcvAx:bw59ahj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L13)):*

> ```
> TEST_LOG_IN = "./sources/replay_log/test_log_in.log"
> TEST_LOG_OUT = "./sources/replay_log/test_log_out.log"
> ```
Doing any more file I/O from unit tests than reading test data feels weird, the filesystem is very 'messy' in that there a ton of ways for things to go wrong that unit tests ideally avoid. I think a good move here would be to (in `replay_log.py`) abstract the opening of the file into `main.py` (take a file-like object) so that your tests can pass a `BytesIO` instead.
</blockquote></details>

I don't this level of bulletproofing is very pragmatic. We can rely on the file system. The good first issue perhaps?

BluCodeGH (changes requested):
Looking great, just a couple of nitpicks.

> Reviewed 1 of 4 files at r4, 1 of 2 files at r5, 2 of 3 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-:-Mnxaez-C0PGFxkj3YRT:b-b8qevc)* status: all files reviewed, 3 unresolved discussions (waiting on @DevinLeamy)

___
*[sources/replay_log/main.py, line 71 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVgNFA9bq48B_cwddS:-Mns2CWTDaiK8M1TrxKo:b-mtca23) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L71)):*
<details><summary><i>Previously, DevinLeamy (Devin Leamy) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Looks like you never actually call the validation function you made lol.
___
*[sources/replay_log/main.py, line 11 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns2Q9y1VhHndtxboxH:-Mns2Q9y1VhHndtxboxI:biajzbd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L11)):*
> ```Python
> 
> 
> def valid_replay_speed(n):
> ```

This being a function is a bit weird for a couple of reasons. Firstly, it's a single comparison that is used exactly once so it's not like it's doing much in terms of reducing code duplication or splitting up code for readability. Secondly, because it's a function it's theoretically usable by other code and so should raise an exception on an error, which you do. However, people who don't understand python don't know where to look in an exception traceback, so when communicating errors to the user it's always better to print a nice message and then `sys.exit(1)`. It's bad practice to call `sys.exit` from a function that could be used elsewhere, so overall I think this check would be cleaner to keep in the main flow.
___
*[sources/replay_log/main.py, line 43 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns3ceOCgV8spGqTBJU:-Mns3ceOCgV8spGqTBJV:b-t2ucw3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L43)):*
> ```Python
> 
>     if len(log_files) == 0:
>         raise Exception("no global log files to replay")
> ```

See above comment on throwing exceptions to the user, better to `return None` like you did but add a check+print+exit in the main flow.
___
*[sources/replay_log/main.py, line 51 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns1oSa3OgcqOmn2dUf:-Mns1oSa3OgcqOmn2dUg:b-sxwas7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L51)):*
> ```
>         if selection == 'R' or (selection.isdigit() and 0 <= int(selection) <= len(log_files)):
>             selection = 0 if selection == 'R' else int(selection)
> ```
Would be more readable to split this into two checks, one for `R` and the other for a number.
___
*[sources/replay_log/test_replay_log.py, line 13 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVqqog3TJgZg8u_xTP:-MnxamhJ1hcKV9E28fq8:bar234y) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L13)):*
<details><summary><i>Previously, DevinLeamy (Devin Leamy) wrote…</i></summary><blockquote>

I don't this level of bulletproofing is very pragmatic. We can rely on the file system. The good first issue perhaps?
</blockquote></details>

Having this sort of shared state between tests still makes me very apprehensive, for example all of these tests break if we run tests in parallel at all. This is an easy way to have buggy tests where their success depends on something like the order they are run in. That being said I'll let it slide in the interest of getting this PR in.

DevinLeamy (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34)* status: 5 of 6 files reviewed, 3 unresolved discussions (waiting on @BluCodeGH and @DevinLeamy)

___
*[sources/replay_log/main.py, line 71 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVgNFA9bq48B_cwddS:-Mo1zk-ADBhD8Ss3o_xC:btg913d) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/main.py#L71)):*

Looks like you never actually call the validation function you made lol.
</blockquote></details>

```
parser.add_argument('--replay_speed', '-r', default=1, type=valid_replay_speed,
                        help="replay speed of log, must be greater than zero (default: 1)")
```
___
*[sources/replay_log/main.py, line 43 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns3ceOCgV8spGqTBJU:-Mo23oB99H-W9o-_KWDm:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L43)):*

See above comment on throwing exceptions to the user, better to `return None` like you did but add a check+print+exit in the main flow.
</blockquote></details>

Done.
___
*[sources/replay_log/main.py, line 51 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns1oSa3OgcqOmn2dUf:-Mo22O968qF1BBUpb-Uj:b-m28wvg) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L51)):*

> ```
>         if selection == 'R' or (selection.isdigit() and 0 <= int(selection) <= len(log_files)):
>             selection = 0 if selection == 'R' else int(selection)
> ```
Would be more readable to split this into two checks, one for `R` and the other for a number.
</blockquote></details>

I didn't like the double break here. 
```
        if selection == 'R':
            selection = 0
            break
        elif selection.isdigit() and 0 <= int(selection) < len(log_files):
            selection = int(selection)
            break
        else:
         ....
```
and I didn't want to use to a try-catch. If you have an alternative approach in mind I'm down for that. Otherwise, I think this way is nicer tbh.
___
*[sources/replay_log/test_replay_log.py, line 13 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-MmVqqog3TJgZg8u_xTP:-Mo2-DChAseZQdSP-EiA:b-gy6gkl) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fbe446f1ebb0130d85cdb4ac2bd1af9902da74d4/sources/replay_log/test_replay_log.py#L13)):*

Having this sort of shared state between tests still makes me very apprehensive, for example all of these tests break if we run tests in parallel at all. This is an easy way to have buggy tests where their success depends on something like the order they are run in. That being said I'll let it slide in the interest of getting this PR in.
</blockquote></details>

I read into this a little bit and you're 100% right. Messing with file I/O is typically a no-go - people take their unit testing seriously lol. Once this is merged I make it an issue on click up.

BluCodeGH (approved):
:lgtm:, make sure to create that clickup task for fixing the tests.

> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-:-Mo744qd2G-0G3kQ8I2C:b-gr7zii)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @DevinLeamy)

___
*[sources/replay_log/main.py, line 51 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/34#-Mns1oSa3OgcqOmn2dUf:-Mo73uZK6FA6l-6DyJ_E:bw8ib66) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/cc6cc3237b1fff2b1cbae173adbf58c58d2c2557/sources/replay_log/main.py#L51)):*
<details><summary><i>Previously, DevinLeamy (Devin Leamy) wrote…</i></summary><blockquote>

I didn't like the double break here. 
```
        if selection == 'R':
            selection = 0
            break
        elif selection.isdigit() and 0 <= int(selection) < len(log_files):
            selection = int(selection)
            break
        else:
         ....
```
and I didn't want to use to a try-catch. If you have an alternative approach in mind I'm down for that. Otherwise, I think this way is nicer tbh.
</blockquote></details>

Yeah I guess that's personal preference, I find the double break more readable.

PR #45, opened 2022-03-14
Tick counter

PCModeActivate:
A wrapper for tick counting purposes; currently implemented on Plot FPS as demo

---
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)


BluCodeGH (changes requested):
Looks good, sorry for the big review just lots of little things. This could also use some unit tests (please keep the time durations short so the unit tests run fast, <1s should be achievable).

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-:-MyAEknVDsco8Auwf34Q:byexk5p)* status: all files reviewed, 12 unresolved discussions (waiting on @PCModeActivate)

___
*[sinks/globallog/tick_counter.py, line 5 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0QBZADyXS8AyDycp:-MyA0QBZADyXS8AyDycq:b-1koxd5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L5)):*
> ```Python
> 
> class TickCounter:
> ```

This class could use a docstring describing what it does.
___
*[sinks/globallog/tick_counter.py, line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA-tZl14J8pHgoITP2:-MyA-tZl14J8pHgoITP3:b-j9fq4y) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L6)):*
> ```Python
> class TickCounter:
> 
>     # avgtick is an integer noting the running average, set to 0 instanatenous rate is desired
> ```

`running_average_length` would be a more descriptive variable name here. Also no need to have any special conditions for if it's zero, set it to 1 and you get the instantaneous rate automatically.
___
*[sinks/globallog/tick_counter.py, line 11 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0BXK1KGc1gLv6xb6:-MyA0BXK1KGc1gLv6xb7:b-6066tb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L11)):*
> ```Python
>         self.count = 0
>         self.last_timestamp = time.time()
>         self.last_tps = 0
> ```

`last_timestamp` is unnecessary since we don't need a special instantaneous case. It's also probably easier to only calculate the tps when asked in `tps()` so we don't need a variable for it.
___
*[sinks/globallog/tick_counter.py, line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0Y5ICS5rb8bkeozm:-MyA0Y5ICS5rb8bkeozn:b-ievplz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L12)):*
> ```Python
>         self.last_timestamp = time.time()
>         self.last_tps = 0
>         self.timestamp_list = [time.time()]
> ```

This class being instantiated isn't necessarily equivalent to a `tick()`, better to instantiate as empty and have `tps()` check and return zero if the length is zero.
___
*[sinks/globallog/tick_counter.py, line 20 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0ptd2cZsgniAufAd:-MyA0ptd2cZsgniAufAe:b-czlll7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L20)):*
> ```Python
>         if self.avgtick > 0:
>             self.timestamp_list.append(time.time())
>             if self.count >= self.avgtick:
> ```

IMO `if len(self.timestamp_list) > self.running_average_length` is more clear about what is going on here. Although the count is equal to the length until we start popping that's not immediately clear (also I think this method might introduce some off by one errors).
___
*[sinks/globallog/tick_counter.py, line 31 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1ITZ4sefIkGHpNxR:-MyA1ITZ4sefIkGHpNxS:bolnpyw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L31)):*
> ```Python
>         self.count += 1
> 
>     def tps(self):
> ```

Since the running average is based on a number of ticks rather than a duration `tick_rate` would be a better name for this function.
___
*[sinks/plot/plot.py, line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1QhQ4BuiNAvzaGz_:-MyA1QhQ4BuiNAvzaGza:b-b8mfmz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L15)):*
> ```Python
> from parsers import Parser
> 
> from sinks.globallog.tick_counter import TickCounter
> ```

The tick counter is a generally useful utility, no reason it should live in globallog (especially given that the code in globallog doesn't even use it lol). Probably better to create `omnibus/util.py` so we can go `from omnibus.util import TickCounter`.
___
*[sinks/plot/plot.py, line 48 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1suB7s5zPddTAlU7:-MyA1suB7s5zPddTAlU8:b8oaafs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L48)):*
> ```Python
>             self.win.addItem(plot.plot, i // columns, i % columns)
> 
>         # adding a viewbox with a textItem in it masquerading as a graph
> ```

Nit, correct tense is `add`
___
*[sinks/plot/plot.py, line 49 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyABLcr4GuZzUKzv_XB:-MyABLcr4GuZzUKzv_XC:bhloh5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L49)):*
> ```Python
> 
>         # adding a viewbox with a textItem in it masquerading as a graph
>         self.textvb = self.win.addViewBox(col=columns - 1, row=len(series) % columns)
> ```

Looks like by default you can pan and zoom a viewbox, best to set `enableMenu=False, enableMouse=False` in this constructor.
___
*[sinks/plot/plot.py, line 52 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1wLRDtGiFGUZ5kPL:-MyA1wLRDtGiFGUZ5kPM:b-y7w5tj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L52)):*
> ```Python
>         self.txitem = TextItem("", (255, 255, 255))
>         self.textvb.autoRange()
>         self.txitem.setPos(0.25, 0.6)  # Centralize the Text
> ```

Nit, correct word is "center" (lowercase)
___
*[sinks/plot/plot.py, line 52 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyABu-iAkpxnIqC8NDA:-MyABu-iAkpxnIqC8NDB:b1u6vcv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L52)):*
> ```Python
>         self.txitem = TextItem("", (255, 255, 255))
>         self.textvb.autoRange()
>         self.txitem.setPos(0.25, 0.6)  # Centralize the Text
> ```

This looks like you've manually found the position where the text is centered. See the "another possible solution" from https://stackoverflow.com/questions/46017022/textitem-added-to-viewbox-disappeare-after-a-couple-of-seconds, looks like you can add `anchor=(0.5, 0.5)` to the text object constructor and then set the position to `(0.5, 0.5)`. 
___
*[sinks/plot/plot.py, line 62 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyAEWYaDBZN5dgY_RjA:-MyAEWYaDBZN5dgY_RjB:b-g4ar62) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L62)):*
> ```Python
>     def update(self):
>         self.counter.tick()
>         fps = self.counter.tps()
> ```

We can move this calculation inside the if statement.

PCModeActivate (commented):
Added 2 tests, 1 second each; not sure why I couldnt run pytests on my own end but should be good to go, as tested with python unittest locally.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-:-My_mESm3qkUbGJr7_Mw:bbcw9uc)* status: 0 of 4 files reviewed, 9 unresolved discussions (waiting on @BluCodeGH)

___
*[sinks/plot/plot.py, line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1QhQ4BuiNAvzaGz_:-Mya1jeO2hDwrgoXbxnA:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L15)):*

The tick counter is a generally useful utility, no reason it should live in globallog (especially given that the code in globallog doesn't even use it lol). Probably better to create `omnibus/util.py` so we can go `from omnibus.util import TickCounter`.
</blockquote></details>

Done.
___
*[sinks/plot/plot.py, line 49 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyABLcr4GuZzUKzv_XB:-Mya1msiCDP2F_Czl5tC:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L49)):*

Looks like by default you can pan and zoom a viewbox, best to set `enableMenu=False, enableMouse=False` in this constructor.
</blockquote></details>

Done.
___
*[sinks/plot/plot.py, line 52 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyABu-iAkpxnIqC8NDA:-Mya1oB70b5r3RVAu-49:b-ml1g9m) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/plot/plot.py#L52)):*

This looks like you've manually found the position where the text is centered. See the "another possible solution" from https://stackoverflow.com/questions/46017022/textitem-added-to-viewbox-disappeare-after-a-couple-of-seconds, looks like you can add `anchor=(0.5, 0.5)` to the text object constructor and then set the position to `(0.5, 0.5)`. 
</blockquote></details>

Yeah Im not sure why 0.5,0.5 wasnt centered correctly; did actually come across that post earlier, but found that 0.5,0.5 was too far left.
___
*[sinks/globallog/tick_counter.py, line 5 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0QBZADyXS8AyDycp:-Mya2BUC9Vo8Ye93iHvB:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L5)):*

This class could use a docstring describing what it does.
</blockquote></details>

Done.
___
*[sinks/globallog/tick_counter.py, line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA-tZl14J8pHgoITP2:-Mya29pe6TqRe83wihJQ:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L6)):*

`running_average_length` would be a more descriptive variable name here. Also no need to have any special conditions for if it's zero, set it to 1 and you get the instantaneous rate automatically.
</blockquote></details>

Done.
___
*[sinks/globallog/tick_counter.py, line 11 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0BXK1KGc1gLv6xb6:-Mya2AAH8T_q5cAVcHXV:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L11)):*

`last_timestamp` is unnecessary since we don't need a special instantaneous case. It's also probably easier to only calculate the tps when asked in `tps()` so we don't need a variable for it.
</blockquote></details>

Done.
___
*[sinks/globallog/tick_counter.py, line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0Y5ICS5rb8bkeozm:-Mya2Br-D9xIuSChY7vp:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L12)):*

This class being instantiated isn't necessarily equivalent to a `tick()`, better to instantiate as empty and have `tps()` check and return zero if the length is zero.
</blockquote></details>

Done.
___
*[sinks/globallog/tick_counter.py, line 20 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA0ptd2cZsgniAufAd:-Mya2DDj10gIijujf-2F:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L20)):*

IMO `if len(self.timestamp_list) > self.running_average_length` is more clear about what is going on here. Although the count is equal to the length until we start popping that's not immediately clear (also I think this method might introduce some off by one errors).
</blockquote></details>

Done.
___
*[sinks/globallog/tick_counter.py, line 31 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-MyA1ITZ4sefIkGHpNxR:-Mya2DZI44J5dIk8p-Ui:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9474e0ba1ffa153e486d9eae98f66f30d95a1dc7/sinks/globallog/tick_counter.py#L31)):*

Since the running average is based on a number of ticks rather than a duration `tick_rate` would be a better name for this function.
</blockquote></details>

Done.

PCModeActivate (commented):
Never mind, pytest works now.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-:-Mya61fX8aIVtY2HrCTA:b-bw17gk)* status: 0 of 4 files reviewed, 9 unresolved discussions (waiting on @BluCodeGH)

BluCodeGH (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: all files reviewed, 2 unresolved discussions (waiting on @PCModeActivate)

___
*[omnibus/tick_counter_test.py, line 11 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd1kG44uRFsY0cys-t:-Myd1kG44uRFsY0cys-u:b-2fg394) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/tick_counter_test.py#L11)):*
> ```Python
>         t = TickCounter(2)
>         t.tick()
>         time.sleep(1)
> ```

If we shorten the test times to say 0.1s total do they still work properly? A full second feels a bit unnecessary given the speed of modern computers.
___
*[omnibus/util.py, line 28 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd06PQ-S9Kx8dCeI3u:-Myd06PQ-S9Kx8dCeI3v:br8rf1u) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/util.py#L28)):*
> ```Python
>         if len(self.timestamp_list) > 1:
>             self.rate = (len(self.timestamp_list)-1) / \
>                 (self.timestamp_list[-1] - self.timestamp_list[0])
> ```

Hmm, consider what would happen if we suddenly stopped calling `tick()`. `self.rate` would never get updated and so the value would just stay constant. Also with this approach even if we did see a `tick()` after a long period of nothing, the measurement would still factor in `n - 1` ticks long ago and not give a very correct answer. I'm thinking we should maybe switch to the "proper" way of doing this which is to average over a fixed time interval rather than a fixed number of ticks. You'd have to switch the `pop` if statement to a while loop that pops until the remaining times in the list are within the set duration, and then have a duplicate of that loop along with the calculation logic in `tick_rate()`.

PCModeActivate (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @BluCodeGH)

___
*[omnibus/tick_counter_test.py, line 11 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd1kG44uRFsY0cys-t:-MzCQO6x8-mC7KwubmV8:b7zeg8v) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/tick_counter_test.py#L11)):*

If we shorten the test times to say 0.1s total do they still work properly? A full second feels a bit unnecessary given the speed of modern computers.
</blockquote></details>

For the instantaneous test, the time.time() method has +0.1 second margin, so 0.1 second instantaneous is actually not quite accurate (1/0.11 is 9.09); this is not a problem unique to our program, since a sample program like "time.time() time.sleep(0.1) time.time()" will yield a 0.11 second difference when printed, so I'm not quite sure how to will counter this problem, if we need to do anything about it at all.
___
*[omnibus/util.py, line 28 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd06PQ-S9Kx8dCeI3u:-MzCeAum3sLqiB9rItSx:boni53t) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/util.py#L28)):*

Hmm, consider what would happen if we suddenly stopped calling `tick()`. `self.rate` would never get updated and so the value would just stay constant. Also with this approach even if we did see a `tick()` after a long period of nothing, the measurement would still factor in `n - 1` ticks long ago and not give a very correct answer. I'm thinking we should maybe switch to the "proper" way of doing this which is to average over a fixed time interval rather than a fixed number of ticks. You'd have to switch the `pop` if statement to a while loop that pops until the remaining times in the list are within the set duration, and then have a duplicate of that loop along with the calculation logic in `tick_rate()`.
</blockquote></details>

This I think we need to take a closer look at; since with this approach tick counter be running in the background even when no ticks are counted, I think we might need to consider giving the user an option to turn on and off the tick counter at their discretion when using omnibus?

Since we're running tick counter in the background now, we also need to either use event scheduling to be called at regular time intervals to take care of tick counting and not disrupt overall program flow.

PCModeActivate (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @BluCodeGH)

___
*[omnibus/tick_counter_test.py, line 11 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd1kG44uRFsY0cys-t:-MzCh3ke33S3MHV7pmM2:b-u9rjt0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/tick_counter_test.py#L11)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

For the instantaneous test, the time.time() method has +0.1 second margin, so 0.1 second instantaneous is actually not quite accurate (1/0.11 is 9.09); this is not a problem unique to our program, since a sample program like "time.time() time.sleep(0.1) time.time()" will yield a 0.11 second difference when printed, so I'm not quite sure how to will counter this problem, if we need to do anything about it at all.
</blockquote></details>

Also weird how github action test passed for the 0.1s setting but my local pytest failed

BluCodeGH (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: all files reviewed, 1 unresolved discussion (waiting on @PCModeActivate)

___
*[omnibus/tick_counter_test.py, line 11 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd1kG44uRFsY0cys-t:-MzbL6lBBj4KJK9Xv3-c:b-4gfmy9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/tick_counter_test.py#L11)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

Also weird how github action test passed for the 0.1s setting but my local pytest failed
</blockquote></details>

Yeah makes sense, I guess it's tricky to make good tests that rely on timing.
___
*[omnibus/util.py, line 28 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd06PQ-S9Kx8dCeI3u:-MzbN8Uj4l3AJG9n95oT:bkhnwvs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/util.py#L28)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

This I think we need to take a closer look at; since with this approach tick counter be running in the background even when no ticks are counted, I think we might need to consider giving the user an option to turn on and off the tick counter at their discretion when using omnibus?

Since we're running tick counter in the background now, we also need to either use event scheduling to be called at regular time intervals to take care of tick counting and not disrupt overall program flow.
</blockquote></details>

We don't need to do any background processing, just some up-front processing when they ask for the rate. See https://gist.github.com/BluCodeGH/301e03a670446d6ead28ea248b038f55.

PCModeActivate (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: all files reviewed, 1 unresolved discussion (waiting on @BluCodeGH)

___
*[omnibus/util.py, line 28 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45#-Myd06PQ-S9Kx8dCeI3u:-MzfGPGqDRh8uUc3Gnap:bksfeg1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4411c16763e6cae969af3322a6ac258291781906/omnibus/util.py#L28)):*

We don't need to do any background processing, just some up-front processing when they ask for the rate. See https://gist.github.com/BluCodeGH/301e03a670446d6ead28ea248b038f55.
</blockquote></details>

That gist doesn't really provide a "proper" way of doing FPS, but if we are okay with this solution to simply "prune" older timestamps, then I'm more than happy to implement it, since it's less complex to do anyways :)

aidenfoxivey (approved):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: all files reviewed, 1 unresolved discussion (waiting on @BluCodeGH)

BluCodeGH (approved):

BluCodeGH (approved):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/45)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @PCModeActivate)

PR #60, opened 2022-09-19
Can table

KavinSatheeskumar:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60)


KavinSatheeskumar (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60)* status: 0 of 23 files reviewed, 11 unresolved discussions (waiting on @aprilove, @BluCodeGH, @dwkletke, and @PCModeActivate)

___
*[`sinks/dashboard/can_display.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCQ6hGeDP_z3ycVWdUu:-NCQ6hGf3H2m7pIESk7S:bqmjo6z) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/410ae28faf4b96f30e6b0f8a45a962011a1fb74b/sinks/dashboard/can_display.py#L10)):*
> ```Python
> from parsers import CanDisplayParser, BOARD_NAME_LIST
> 
> # --------------CONSTANTS---------------
> ```

First, if we are going to have this many constants, they should be in another file. Secondly, we need to reduce the dependency on these constants. During one of our tests, a new board ID ended up absolutely crashing the entire program. We should get someone to do something about the dependency on hard coded constants.
___
*[`sinks/dashboard/can_display.py` line 56 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCQ8ECz9vovM9Ch5Nip:-NCQ8ECz9vovM9Ch5Niq:bmg6chh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/410ae28faf4b96f30e6b0f8a45a962011a1fb74b/sinks/dashboard/can_display.py#L56)):*
> ```Python
> class CanDisplayDashItem (DashboardItem):
>     """
>     Display for CAN messages.
> ```

A more comprehensive description is needed. Looking through this code, its hard to find out what the dashitem is supposed to look like. Eg. What are the buttons supposed to do? What is the layout of this thing? Etc.
___
*[`sinks/dashboard/can_display.py` line 54 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq5_lE6w7SaF_x2Tc2:-NCq5_lE6w7SaF_x2Tc3:b-k3fda1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/can_display.py#L54)):*
> ```Python
> 
> 
> class CanDisplayDashItem (DashboardItem):
> ```

Now that we have the can-table, do we even need this entire Widget?
___
*[`sinks/dashboard/can_display.py` line 80 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq2g2Y9P9sK_XADxMy:-NCq2g2Y9P9sK_XADxMz:b2qgmqh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/can_display.py#L80)):*
> ```Python
>         # add the above to the local layout (right for buttons/labels, left for terminals)
>         for index, value in enumerate(list(BOARD_DATA.keys())):
>             self.canNodes[index] = CanNodeWidgetDashItem(value)
> ```

We have a dashitem nested inside a dashitem. That is not how these things were intended to work.

The real thing to do is to seperate out the series_observer class from the dash_item class so we aren't forced to do things like this.
___
*[`sinks/dashboard/can_display.py` line 149 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq3MTf8YSQp2RolEMy:-NCq3MTf8YSQp2RolEMz:b-81z9ox) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/can_display.py#L149)):*
> ```Python
> class CanNodeWidgetDashItem(DashboardItem):
>     """
>     Display for CAN messages.
> ```

This class is meant to represent a terminal, but its name gives no indication of that and the docstring is identical to the other class.
___
*[`sinks/dashboard/can_message_table.py` line 6 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq1Zrl5po8pM4YvGVV:-NCq1Zrl5po8pM4YvGVW:bcylv40) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/can_message_table.py#L6)):*
> ```Python
> from parsers import CanDisplayParser
> 
> CAN_MSG_TYPES = ["GENERAL_CMD",
> ```

Same issue as above for these hard coded constants, we need to have our code not depend on the setup for the hardware
___
*[`sinks/dashboard/dashboard.py` line 31 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq5pXHDWLpK8TbvSEZ:-NCq5pXHDWLpK8TbvSE_:b2zh6de) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/dashboard.py#L31)):*
> ```Python
> class Dashboard(QtWidgets.QWidget):
>     """
>     Displays a grid of plots in a window
> ```

this is pedantic, but grid of widgets
___
*[`sinks/dashboard/dashboard.py` line 42 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq70TE-tCqT_Vdv7F0:-NCq70TE-tCqT_Vdv7F1:b-52vwmr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/dashboard.py#L42)):*
> ```Python
> 
>         # The file from which the dashboard is loaded
>         self.filename = "savefile.sav"
> ```

We need to do something better with the file name, I feel like it should start empty (IE we should always load nothing) and then people must manually open a file.

I think that starting with this as `None` does the job
___
*[`sinks/dashboard/dashboard.py` line 115 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq7QcrACWbRjO-DwHV:-NCq7QcrACWbRjO-DwHW:b-coxloj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/dashboard.py#L115)):*
> ```Python
>         # registered with any series
>         for dock in self.docks:
>             item = dock.widgets[0]
> ```

Add a comment here say that each dock only contains 1 widget, meaning we are just grabbing it.

Note that this abuses the implementation of pyqtgraph, that should also be noted and the specific implementation should be cited.
___
*[`sinks/dashboard/dashboard.py` line 191 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq9L8O6bCWxvUIrvdO:-NCq9L8P2h4qDQzDlG3J:b5ubeh5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/dashboard.py#L191)):*
> ```Python
>         dock = Dock(f"{len(self.docks)}", closable=True)
> 
>         # Bit of a sussy baka, but this is the only way we can really get control over
> ```

Remove this comment, not nearly as sussy of a baka as I initially thought
___
*[`sinks/dashboard/dashboard.py` line 229 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/60#-NCq9m4l4ydwxKaG3gGN:-NCq9m4l4ydwxKaG3gGO:b695ksj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/6e359b5352aea3e4e0bfe605a72cfa811cd3b248/sinks/dashboard/dashboard.py#L229)):*
> ```Python
>         # If the filename entered is not valid
>         # this exhibits the behaviour of creating
>         # a new one
> ```

No it dosen't, it just adds it to the cache, and only if the user saves, does it create a new file name. We should really check to see if the file is valid or not before attempting to load from it. (Just like how we check to see if its none above)

PR #62, opened 2022-10-31
Put dashItems in their own folders

Darth-Raazi:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62)


BluCodeGH (changes requested):
You didn't delete the files in their old locations?

Darth-Raazi:
Weird... I did delete them, but that didn't get committed. Does 'git add *' not add deleted files? Thanks for catching that!

BluCodeGH (changes requested):
Yeah, so `*` enumerates the files and folders in the current directory. That means it won't catch anything that's been deleted from the current directory, since they are no longer in the current director and thus not enumerated. You'll notice that if you delete something in a subfolder `*` will catch it, and that's because `*` will enumerate the name of that subfolder and then git will recurse the subfolder itself and git is smart enough to catch deleted files. If you want to use the smart git behavior everywhere you can use `git add .`, where `.` refers to the current directory, and then git does its smart recursion everywhere instead of just in subfolders.

Overall, I think the idea here was to create a single `items` folder for all the dashboard items. Moving each item into its own folder doesn't really do anything to clean up the file structure, it just makes it more complicated! Folders should be used to group related code, like the related dashboarditems.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-:-NFgQj8R3L1GpUINgmaO:b-x94q42)* status: all files reviewed, 3 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/can_message_table/can_message_table.py` line 7 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-NFgRJm7-46upI7JzP2q:-NFgRJm7-46upI7JzP2r:bcl6lqc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/86a3359fb8927920aa6fceb8e253673ccc5b6ceb/sinks/dashboard/can_message_table/can_message_table.py#L7)):*
> ```Python
> current = os.path.dirname(os.path.realpath(__file__))
> parent = os.path.dirname(current)
> sys.path.append(parent)
> ```

This is hacky and not good practice. Instead, since omnibus is a module, you should be able to do `from omnibus.sinks.dashboard.parsers import etc`.
___
*[`sinks/dashboard/can_message_table/can_message_table.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-NFgSQU364HAfjdfm38c:-NFgSQU364HAfjdfm38d:b1izirs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/86a3359fb8927920aa6fceb8e253673ccc5b6ceb/sinks/dashboard/can_message_table/can_message_table.py#L10)):*
> ```Python
> 
> from pyqtgraph.Qt import QtCore, QtGui, QtWidgets
> from dashboarditem import DashboardItem
> ```

IMO we should move the generic dashboarditem base class into the common `items/` folder too, simplifies the dependency structure a bit and again groups related code.
___
*[`sinks/dashboard/plot_dash_item/plot_dash_item.py` line 7 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-NFgRoUtCdmtL4tG-Ufp:-NFgRoUtCdmtL4tG-Ufq:b-q64pkt) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/86a3359fb8927920aa6fceb8e253673ccc5b6ceb/sinks/dashboard/plot_dash_item/plot_dash_item.py#L7)):*
> ```Python
> current = os.path.dirname(os.path.realpath(__file__))
> parent = os.path.dirname(current)
> sys.path.append(parent)
> ```

See above.

Darth-Raazi:
Oh I see. I thought `*` was a wildcard statement and just added all your changes. Thanks!

Turns out that `sinks` isn't an object in omnibus but a separate object itself so I changed the import path to `from sinks.dashboard.items... import ...`.

I checked with Kavin and he said you made a good point about grouping the items, so I created an `items` folder and put `can_message_table`, `plot_dash_item` and `dashboard_item` in there.

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH and @Darth-Raazi)

___
*[`sinks/dashboard/can_display.py` line 5 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-NFicd5G-9BdgGaH2XGB:-NFicd5G-9BdgGaH2XGC:bcb7v0j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/90ea1d0b234fd09d368dd9d0e29c26c86fda9efe/sinks/dashboard/can_display.py#L5)):*
> ```Python
> from pyqtgraph.Qt import QtCore, QtGui, QtWidgets
> 
> from items.dashboard_item.dashboard_item import DashboardItem
> ```

I know this is a tad pedantic, but I would prefer it if the `dashboard_item.py` file was not put into its own folder, but just existed within the top level of this directory. The Original point of putting it there in the first place was clean up the import path, and having it here adds this extra bit to the path.

So

items =>
   dashboard_item.py
   [... other folders ...]
___
*[`sinks/dashboard/can_message_table/can_message_table.py` line 7 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-NFgRJm7-46upI7JzP2q:-NFideIc5rzi3p3roPlv:b-k985xa) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/86a3359fb8927920aa6fceb8e253673ccc5b6ceb/sinks/dashboard/can_message_table/can_message_table.py#L7)):*

This is hacky and not good practice. Instead, since omnibus is a module, you should be able to do `from omnibus.sinks.dashboard.parsers import etc`.
</blockquote></details>

Now that you have fixed this, I think you forgot to remove the `sys` import here.

BluCodeGH (approved):
:lgtm:

> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-:-NFlCqxMCCymdnzRkBaf:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Darth-Raazi and @KavinSatheeskumar)

KavinSatheeskumar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/62#-:-NFlIIgiE5zc-16ZlGV7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Darth-Raazi)

PR #71, opened 2022-11-09
Publisher and Subscriber Abstraction (Observer Pattern)

PCModeActivate:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)


BluCodeGH (changes requested):
I may be misremembering but I believe this is a different architecture than we last agreed on. The architecture I had in mind was a a single publisher class that used strings to differentiate the different "streams" that could be subscribed to, as opposed to the current implementation that uses class instances for that differentiation.

The main benefit of the string-based approach is that anything can subscribe to a stream knowing nothing more than its stream ID, which allows for functionality like having a dashboarditem subscribed to a stream that hasn't had any data yet. It also generally simplifies the implementation, for example there is no need for `SeriesDefaultDict` since the parsers can pass the hardcoded string ID directly to the single publisher class. This also lets you get rid of things like `Parser.get_series` (called by dashitems on load to get a handle of their series) as instead those dashitems can subscribe to the saved stream ID, reducing coupling (dashitems don't need to know about parsers).

I see this being implemented as a class only so that the relevant methods are grouped, otherwise I would suggest global functions operating on global state. Since there is no need for multiple independent states, we can use static methods and operate on a single state stored in the class. I would structure the state as a dictionary mapping stream IDs (strings) to lists of observers, so when we get new data we iterate through the respective list and notify the observers.

I would also switch how data is passed to the publisher. With the (upcoming) rewrite of the parsers, we will want the ability for one parser to easily add multiple datapoints to a variety of streams. In that light I think the publish method should take in a timestamp, which will be added to all the data, and a dictionary mapping stream IDs to datapoints. The publish method would then:

```python
@staticmethod
def publish(timestamp, datapoints):
    for stream, value in datapoints.items():
        for observer in Publisher.stream_observer_map_internal_state_whatever[stream]:
            observer.notify(timestamp, value)
```

Lastly this PR still has a _lot_ of code referring to "series" (files, function names, etc), that should be fixed in the next revision. To summarize the nomenclature in a sentence:  "An Observer subscribes to a Stream of data on the Publisher, which is sent said data by Parsers".

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NGVKA09ChEhqImqA7EE:b-5b3k9g)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @KavinSatheeskumar)

KavinSatheeskumar (changes requested):
SO, when I was initially speaking with you the picture I had in mind was this.

Some kind of global dictionary with key's as ID's and values as publishers. Very similar to the way the current code base works. This is similar to what is in the code, currently. I think I now understand that what you wanted was the class publisher to take the role of that dictionary. I agree, that is probably better as it makes querying, subscribing and adding "streams" (although, the better term is "mailing lists" ;P). The one thing it makes more annoying is the update but it won't be much more annoying and its a fair trade off for the rest being so simple. Therefore, @pcmodeactivate , Imma have to side with Jack. Sorry I am forcing you to re-write this.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NGZjO2h8tY6UJULk89j:b-vzk7xc)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @BluCodeGH and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 9 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZMnouApVBYX9-Ze_J:-NGZMnovAtO1OFEJWwi3:b-vl049i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/parsers.py#L9)):*
> ```Python
> 
> 
> class SeriesDefaultDict(defaultdict):
> ```

One thing I do agree with Jack on is that this scheme with SeriesDefaultDict is kinda clunky and should be worked around. 
___
*[`sinks/dashboard/series.py` line 11 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZNV0j6yoT6dnMSeIi:-NGZNV0j6yoT6dnMSeIj:bbl8yq5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/series.py#L11)):*
> ```Python
>     def __init__ (self, name, time_rollover = False):
>         self.name = name
>         self.callback = None
> ```

We never use self.callback
___
*[`sinks/dashboard/series.py` line 28 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZkdMvBRCSC-8jAPB4:-NGZkdMvBRCSC-8jAPB5:b-uxyjxm) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/series.py#L28)):*
> ```Python
>         self.observers.remove(obs_to_remove)
> 
>     def notify_observers(self):
> ```

The "add" and "notify observers" function should be rolled into one. The `observer.on_data_update` should take the payload as the argument rather than the entire series. New notify should look like the code snippit below



_Code snippet:_
```Python
def notify_observers(self, payload):
    for observer in self.observers:
        observer.on_data_update(payload)
```
___
*[`sinks/dashboard/items/dashboard_item.py` line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZlROP2zxL4YDPZJrJ:-NGZlROP2zxL4YDPZJrK:b-h68gx0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L6)):*
> ```Python
> 
> 
> class Subscriber():
> ```

The definition of this should be in same file as publisher IMO
___
*[`sinks/dashboard/items/dashboard_item.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZmxLi4xnDEp7TlDcJ:-NGZmxLi4xnDEp7TlDcK:b-xmm9sm) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L10)):*
> ```Python
>         self.subscribed_subjects = []
> 
>     def subscribe_to(self, subject):
> ```

Yeah, should have specified this, but we wanted to make part of the point of this was to make it easier to subscribe to series without a bunch of coupling. 
___
*[`sinks/dashboard/items/dashboard_item.py` line 27 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZn6CkEZ4nMz_6vkMA:-NGZn6CkEZ4nMz_6vkMB:b-4ke5q8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L27)):*
> ```Python
>             subject.remove_observer(self)
> 
>     def on_data_update(self, series):
> ```

should take a new data point, rather than an entire series

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @BluCodeGH and @PCModeActivate)

___
*[`sinks/dashboard/series.py` line 24 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWoQUP7GyINzM2ODVM:-NHWoQUP7GyINzM2ODVN:b-iz37ut) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L24)):*
> ```Python
>             series.remove(observer)
> 
>     def update(self, series, payload):
> ```

just a thought, but maybe we should make update take 4 arguments, self, series, timestamp and payload. Since we always know that we are passing the timestamp, may as well just pass it here

I think we had ought change the on_data_update method accordingly
___
*[`sinks/dashboard/series_test.py` line 13 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWovon1v6UJKi5Jl8m:-NHWovooA03LARDh3ohQ:b-gio28r) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series_test.py#L13)):*
> ```Python
>         monkeypatch.setattr(config, "GRAPH_DURATION", 10)
> 
>     def test_no_downsample(self):
> ```

Just noting that `add` is no longer a method, but also all the tests are so wrong as this point is doesn't matter, just never use pytest
___
*[`sinks/dashboard/items/dashboard_item.py` line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZlROP2zxL4YDPZJrJ:-NHWpBeW8TYVRp-vrpTm:b32wed) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L6)):*

The definition of this should be in same file as publisher IMO
</blockquote></details>

^^
___
*[`sinks/dashboard/items/plot_dash_item.py` line 118 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWnCh82c56TbnD6hY-:-NHWnCh82c56TbnD6hY0:b-m2t2th) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L118)):*
> ```Python
> 
>         # current value readout in the title
>         #if (desc is not ""):
> ```

No random commented out code pls

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: 0 of 10 files reviewed, 11 unresolved discussions (waiting on @BluCodeGH and @PCModeActivate)

___
*[`sinks/dashboard/series.py` line 1 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWp__r0hRpNxwkr6Wa:-NHWp__r0hRpNxwkr6Wb:b1hgirf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L1)):*
> ```Python
> class Publisher:
> ```

It'd be nice if we could have a method that let's us get the list of all series_names that currently exist for plot_dash_item and things
___
*[`sinks/dashboard/series.py` line 9 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpX3LBspd__7UBetp:-NHWpX3LBspd__7UBetq:blckohv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L9)):*
> ```Python
>         self.canserieses = {}
> 
>     def subscribe(self, series, observer):
> ```

maybe call it series_name rather than series
___
*[`sinks/dashboard/items/plot_dash_item.py` line 43 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWplHaAVbNpmtetwlc:-NHWplHaAVbNpmtetwld:b-tx9ign) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L43)):*
> ```Python
>         # prompt the user for them
>         if self.props == None:
>             items = list(publisher.serieses.keys())
> ```

ye, instead of this, just add a method to publisher that returns this array
___
*[`sinks/dashboard/items/plot_dash_item.py` line 68 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpvgv6hLqFsTS12g6:-NHWpvgv6hLqFsTS12g7:bj8ygog) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L68)):*
> ```Python
>         self.layout.addWidget(self.widget, 0, 0)
> 
>     def on_data_update(self, payload):
> ```

See above, but to clarify I want this to take 3 arguments now, self, timestamp and payload, just so we don't have to worry about splitting things up all the time
___
*[`sinks/dashboard/items/plot_dash_item.py` line 95 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWqIXaDa71KGezLW47:-NHWqIXaDa71KGezLW48:b5sfqo8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L95)):*
> ```Python
> 
>         # add the new datapoint to the end of each array, shuffle everything else back
>         self.times[:-1] = self.times[1:]
> ```

I think it might be a bit clearer using append and pop rather than these indexing tricks. Just my opinion tho
___
*[`sinks/dashboard/items/plot_dash_item.py` line 123 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWqnoN73OpLbLGR4wr:-NHWqnoN73OpLbLGR4ws:b-r803tb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L123)):*
> ```Python
>         #else:
>         self.plot.setTitle(
>             f"[{sum(points)/len(points): <4.4f}] [{self.points[-1]}] {self.props}")
> ```

round off {self.points[-1]} pls, it'll just save us time in the future

PCModeActivate (commented):
I've changed it accordingly and added global publisher _on_top_ of the changes to parser from @darth-raazi. So now it should have no parser object (only parser methods) and one global publisher with subscribers listening in on it. Only glaring problem I have seen from testing is that timestamp starts at $1.6 \cdot 10^9$, but that seems to be the raw data from FakeNI, as it appears to be the data being fed into parser. Other than that, functionally both DAQ and CAN should still work as intended (with better performance compared to dashboard branch, surprisingly).

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NH_bmqw4kqEdfsx8NlE:bdbuvay)* status: 0 of 12 files reviewed, 11 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`sinks/dashboard/series_test.py` line 13 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWovon1v6UJKi5Jl8m:-NH_ZHXt7L1R7tH4HEDQ:bn42wad) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series_test.py#L13)):*

Just noting that `add` is no longer a method, but also all the tests are so wrong as this point is doesn't matter, just never use pytest
</blockquote></details>

Yep. All the tests are wrong now.
___
*[`sinks/dashboard/items/dashboard_item.py` line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZlROP2zxL4YDPZJrJ:-NH_ZPPrDtusBjAdY0CP:b-afm2vq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L6)):*

^^
</blockquote></details>

Well it is its own thing now.
___
*[`sinks/dashboard/items/dashboard_item.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZmxLi4xnDEp7TlDcJ:-NH_ZuqtDiQxu801dw4Y:b-allzvh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L10)):*

Yeah, should have specified this, but we wanted to make part of the point of this was to make it easier to subscribe to series without a bunch of coupling. 
</blockquote></details>

I think this is the standard way of interacting with observer pattern. It's not necessarily more coupling, because otherwise you'd just have to directly import the global publisher and manually call publisher function, which I think is worse, because this actually is a good enough abstraction that groups together common functionalities imo.
___
*[`sinks/dashboard/items/dashboard_item.py` line 27 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZn6CkEZ4nMz_6vkMA:-NH_ZkB68CiG5FcjWrb9:bjj49q6) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L27)):*

should take a new data point, rather than an entire series
</blockquote></details>

This doesn't do anything because the behaviour of it is defined by subclass.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 43 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWplHaAVbNpmtetwlc:-NH_bh2U1dJMYYohdQFf:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L43)):*

ye, instead of this, just add a method to publisher that returns this array
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 68 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpvgv6hLqFsTS12g6:-NH__xvy-qqKrECykJFA:b-3rdboq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L68)):*

See above, but to clarify I want this to take 3 arguments now, self, timestamp and payload, just so we don't have to worry about splitting things up all the time
</blockquote></details>

No. See above.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 95 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWqIXaDa71KGezLW47:-NH_a1Ye2-7Wawx7Rsrr:b9inq8q) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L95)):*

I think it might be a bit clearer using append and pop rather than these indexing tricks. Just my opinion tho
</blockquote></details>

idk, if I recall correctly Jack wrote this (iirc I wrote with append and pop and Jack changed it to this?). Also this is probably more efficient.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 118 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWnCh82c56TbnD6hY-:-NH_aHJ70gTV19f0jV-j:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L118)):*

No random commented out code pls
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 123 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWqnoN73OpLbLGR4wr:-NH_beTR840L98ULBAGZ:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/items/plot_dash_item.py#L123)):*

round off {self.points[-1]} pls, it'll just save us time in the future
</blockquote></details>

Done.
___
*[`sinks/dashboard/series.py` line 1 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWp__r0hRpNxwkr6Wa:-NH_XgakDHfQsLdvCBNc:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L1)):*

It'd be nice if we could have a method that let's us get the list of all series_names that currently exist for plot_dash_item and things
</blockquote></details>

Done.
___
*[`sinks/dashboard/series.py` line 9 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpX3LBspd__7UBetp:-NH_XaMk6zQSt2xehB9c:brom326) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L9)):*

maybe call it series_name rather than series
</blockquote></details>

Now that series is not a real object series is literally just a name, so nah.
___
*[`sinks/dashboard/series.py` line 24 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWoQUP7GyINzM2ODVM:-NH_YkfS0syEfN4Fzd30:b2hf2n9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L24)):*

just a thought, but maybe we should make update take 4 arguments, self, series, timestamp and payload. Since we always know that we are passing the timestamp, may as well just pass it here

I think we had ought change the on_data_update method accordingly
</blockquote></details>

Payload is [data + timestamp] (for now), I am not sure if CAN uses it the same way but CAN logic takes in a payload directly, so I'd rather not change it, at least as of this PR.

PCModeActivate (commented):
*"data" being fed into the parser -> timestamp fed into the parser
The parsing function is fine, that was just my typo.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NH_dpVU0bdj659JjFdM:b3wv73m)* status: 0 of 12 files reviewed, 10 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: 0 of 12 files reviewed, 5 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 31 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaMYWX8FGzBhmV97MH:-NHaMYWX8FGzBhmV97MI:b-taritr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L31)):*
> ```Python
> 
> 
> def parse(msg_channel, msg_payload):
> ```

This had ought be the last thing in the file
___
*[`sinks/dashboard/items/dashboard_item.py` line 27 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZn6CkEZ4nMz_6vkMA:-NHaU7kABYhmthnr_oEI:b-41c6wo) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L27)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

This doesn't do anything because the behaviour of it is defined by subclass.
</blockquote></details>

We should still change the name of the input value to make the code easier to read
___
*[`sinks/dashboard/publisher.py` line 8 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaQrC-4BKnBoUy0eSl:-NHaQrC0CHfs5A7VAVn0:b-6esafv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L8)):*
> ```Python
>     def __init__(self):
>         self.serieses = {}
>         self.canserieses = {}
> ```

We shouldn't have seperate dictionaries based on types, rather, the values in the dictionary should provide more information about the series.

I am envisioning a brand new series class that looks like this



_Code snippet:_
```Python
class Series:
    observers: List[Subscriber]
    type: Type
```
___
*[`sinks/dashboard/publisher.py` line 11 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaRZ3R2Y7JD7GOAKTQ:-NHaRZ3R2Y7JD7GOAKTR:b-i6wj1l) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L11)):*
> ```Python
> 
>     def get_CAN_series(self):
>         return self.canserieses.keys()
> ```

1. We will only have on series-getting function
2. The return type should be a list of objects, each of which describe a series, (look at the example below)
3. optional argument to filter by type



_Code snippet:_
```Python
class SeriesDescriptor:
    name: str
    type: Type
```
___
*[`sinks/dashboard/series.py` line 9 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpX3LBspd__7UBetp:-NHaSdQzDHRpbHpYZNyW:bfk5sem) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L9)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

Now that series is not a real object series is literally just a name, so nah.
</blockquote></details>

1. the user dosen't know that
2. It now **is** so hah
___
*[`sinks/dashboard/series.py` line 24 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWoQUP7GyINzM2ODVM:-NHaSkgO0f6HdrelqcF5:bqy67xj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L24)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

Payload is [data + timestamp] (for now), I am not sure if CAN uses it the same way but CAN logic takes in a payload directly, so I'd rather not change it, at least as of this PR.
</blockquote></details>

Fair enough

PCModeActivate (commented):
IMPORTANT: This PR will break CAN Parsing, by the fact that CAN is parsed differently from DAQ and our parse() function is not compatible with the different structure of the payload, lest we hard code it. To restructure CAN parsing would be a task worthy of another PR, so what we do exactly with that is up to discussion (e.g. we will need additional logic to handle finding which board to update in the payload, and that requires hard-coding). We could potentially offer a hard-coded solution should it need to be merged into dashboard (if that is what we want), but obviously a revamp for CAN parsing is going to be needed regardless.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NHedD1E-dukXfYsYsPo:b-dt1bvw)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`sinks/dashboard/parsers.py` line 31 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaMYWX8FGzBhmV97MH:-NHea9FhCP0_6j1a4mir:b-29aapu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L31)):*

This had ought be the last thing in the file
</blockquote></details>

It is now.
___
*[`sinks/dashboard/items/dashboard_item.py` line 27 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NGZn6CkEZ4nMz_6vkMA:-NHebrh35iXKuXAMwRV6:b-886o9d) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/69b007081746737a93477803cfed06506c0cd34e/sinks/dashboard/items/dashboard_item.py#L27)):*

We should still change the name of the input value to make the code easier to read
</blockquote></details>

Fine.
___
*[`sinks/dashboard/publisher.py` line 8 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaQrC-4BKnBoUy0eSl:-NHeaLdw-9oGPPtq7FbK:bt5kbrj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L8)):*

We shouldn't have seperate dictionaries based on types, rather, the values in the dictionary should provide more information about the series.

I am envisioning a brand new series class that looks like this
</blockquote></details>

Nope, doesnt work with name collision. We are putting the type as a string at the end of the series name to avoid collision, much like the way we did before.
___
*[`sinks/dashboard/publisher.py` line 11 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaRZ3R2Y7JD7GOAKTQ:-NHeb3Zu9mjSIDko382G:bbyp973) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L11)):*

1. We will only have on series-getting function
2. The return type should be a list of objects, each of which describe a series, (look at the example below)
3. optional argument to filter by type
</blockquote></details>

1. We do have only one series-getting function now.
2. We do have a list of objects now, except it returns the key (str) instead. See above for context on why.
3. It's in there.
___
*[`sinks/dashboard/series.py` line 9 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHWpX3LBspd__7UBetp:-NHebRzr8yfgUmpM4tI7:b-yt75y4) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/d401202e8040cc10f694ee5bb3fd8cf207440db4/sinks/dashboard/series.py#L9)):*

1. the user dosen't know that
2. It now **is** so hah
</blockquote></details>

1. It makes more sense to subscribe to a series, plus, it makes sense to access the dictionary of series by "series in serieses" (lest we change it to "dict of series names").
2. Not anymore, so sucks to be you.

PCModeActivate (commented):
Otherwise DAQ runs fine. If not better than before.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NHeeQTm-UXaMBI21Clv:b-yzo6x5)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

PCModeActivate (commented):
Added hacky fix for CAN in the last two commits. CAN should work now, but it's ugly.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NHf2rpL-Nxrnokbal8w:bgigy3m)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

BluCodeGH (changes requested):
> Reviewed 1 of 6 files at r1, 1 of 6 files at r4, 7 of 10 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: all files reviewed, 13 unresolved discussions (waiting on @Darth-Raazi, @KavinSatheeskumar, and @PCModeActivate)

___
*[`.gitignore` line 29 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbu4h_C7Fn84hfIgy4:-NHbu4h_C7Fn84hfIgy5:b2s2ria) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/.gitignore#L29)):*
> ```ignore
> 
> cscope*
> runtest/*
> ```

What is this?
___
*[`sinks/dashboard/parsers.py` line 7 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbuqej5_hMn6q3taKX:-NHbuqej5_hMn6q3taKY:b37hp9d) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L7)):*
> ```Python
> 
> 
> def register(msg_channels):
> ```

We should change this to be a class decorator so `_func_map` can be a local member.
___
*[`sinks/dashboard/parsers.py` line 14 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbv5-z--flst2P5ZTv:-NHbv5-z--flst2P5ZTw:bexo5ad) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L14)):*
> ```Python
>         for msg_channel in msg_channels:
>             if msg_channel in _func_map:
>                 raise KeyError(f"Duplicate parsers for message type {msg_channel}")
> ```

I think this was mentioned in the other PR but (unlike with parsley where this code came from) we actually want the ability to have multiple parsers for the same message channel. I'd make `_func_map` a dictionary mapping channels to lists of parser functions.
___
*[`sinks/dashboard/parsers.py` line 22 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbvjupBfaPBUzuI1uJ:-NHbvjuq2Kz4-zT_bZt8:b48kdw0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L22)):*
> ```Python
> @register("DAQ")
> def daq_parser(msg_data):
>     timestamp = msg_data["timestamp"]
> ```

Re: the issue with DAQ timestamps, I think it's possible to make a class and decorate one of its static methods so you can use class level variables to store state such as the first time received (like the old parser did). Probably worth investigating and seeing how complex the code ends up as, since storing state is useful.
___
*[`sinks/dashboard/parsers.py` line 31 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaMYWX8FGzBhmV97MH:-NHbvWuRAS88DP9hfChd:bbba0kd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L31)):*

This had ought be the last thing in the file
</blockquote></details>

On the contrary it should be just after `register`, before any of the parsers. Don't want to have to scroll through all of that to see the common logic.
___
*[`sinks/dashboard/parsers.py` line 26 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgLThR7Iu2n84ttX8W:-NHgLThR7Iu2n84ttX8X:bwta0jw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/parsers.py#L26)):*
> ```Python
> 
>     for sensor, data in msg_data["data"].items():
>         parsed_messages.append((sensor + "|DAQ" , timestamp, sum(data)/len(data)))
> ```

Probably better to put "DAQ" first.
___
*[`sinks/dashboard/parsers_test.py` line 32 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbwxnVBbbaGU2-YQEy:-NHbwxnVBbbaGU2-YQEz:b32vmt) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers_test.py#L32)):*
> ```Python
>         # values should be averagd
>         #assert parser.series.get("SENSOR 1").data == [(0, 3)]
>         #assert parser.series.get("SENSOR 2").data == [(0, 5)]
> ```

??
___
*[`sinks/dashboard/items/subscriber.py` line 4 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNX7vC6Zi70BKMILk:-NHgNX7vC6Zi70BKMILl:bj9rv3c) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L4)):*
> ```Python
> 
> 
> class Subscriber():
> ```

I don't see the purpose of this class except for defining the schema of a subscriber, aka`on_data_update`. Both other functions are wrappers which add an unnecessary level of indirection and don't provide any useful functionality.

I'm honestly not even convinced of the value of defining that schema when instead `publisher.subscribe` could take a callback (ie `self.on_data_update`) rather than just an instance of a subscriber. Multiple inheritance is messy and using callbacks lets (for example) a single subscriber handle different subscriptions differently.
___
*[`sinks/dashboard/items/subscriber.py` line 30 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNgn498XfA9Q_s5Gp:-NHgNgn498XfA9Q_s5Gq:by8t4v7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L30)):*
> ```Python
>         as a parameter
>         """
>         pass
> ```

This should `raise NotImplementedError`.
___
*[`sinks/dashboard/publisher.py` line 7 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbx8S7CEleZYq615J1:-NHbx8S7CEleZYq615J2:b-vfji2z) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L7)):*
> ```Python
> 
>     def __init__(self):
>         self.serieses = {}
> ```

`serieses` isn't a word, the plural of series is series. I recommend using the terms "stream" and "streams" instead, since we aren't storing a series of data anymore its just a live stream of it.
___
*[`sinks/dashboard/publisher.py` line 11 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaRZ3R2Y7JD7GOAKTQ:-NHgMa3NCpwbOqmor7yQ:b-2v376i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L11)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

1. We do have only one series-getting function now.
2. We do have a list of objects now, except it returns the key (str) instead. See above for context on why.
3. It's in there.
</blockquote></details>

I don't see the need for any "types" of streams. The CAN table can just subscribe to the "CAN" stream as-is (since there is only the one), and generally in the future things should only subscribe to the explicit streams they are designed to subscribe to. This currently isn't really useful for "dynamic" displays anyway since it just gets called once at startup before we have any idea what streams there actually will be.

PCModeActivate (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: 5 of 12 files reviewed, 11 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`.gitignore` line 29 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbu4h_C7Fn84hfIgy4:-NHkyyvyD2sOjOmzYagv:b3px3j3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/.gitignore#L29)):*

What is this?
</blockquote></details>

Well it's gone now, not something that ought to be global.
___
*[`sinks/dashboard/items/subscriber.py` line 4 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNX7vC6Zi70BKMILk:-NHl-Xh34GNLEqcpEGAj:b113gtg) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L4)):*

I don't see the purpose of this class except for defining the schema of a subscriber, aka`on_data_update`. Both other functions are wrappers which add an unnecessary level of indirection and don't provide any useful functionality.

I'm honestly not even convinced of the value of defining that schema when instead `publisher.subscribe` could take a callback (ie `self.on_data_update`) rather than just an instance of a subscriber. Multiple inheritance is messy and using callbacks lets (for example) a single subscriber handle different subscriptions differently.
</blockquote></details>

I don't think we should be feeding back the `on_data_update` as a callback, since practically every "subscriber" "subclass" (airquotes because subscriber is gone now) should have its own `on_data_update`. It will be called upon update, so it doesn't need to be fed back into the publisher. Subscriber used to be an explicit inheritance structure that was designed to allow an observer to listen in on more than one stream, other than that it just calls `on_data_update`. In any case the observer classes should always have an  `on_data_update`, so it should be fine to leave it to call `on_data_update` from publisher directly.
___
*[`sinks/dashboard/publisher.py` line 7 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbx8S7CEleZYq615J1:-NHl1ghxEC-CQ0I9ju8U:b-pgqi86) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L7)):*

`serieses` isn't a word, the plural of series is series. I recommend using the terms "stream" and "streams" instead, since we aren't storing a series of data anymore its just a live stream of it.
</blockquote></details>

Apparently `serieses` was an old expression for plural for `series` (according to google, /shrug). But still, stream is more apt here.
___
*[`sinks/dashboard/publisher.py` line 11 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaRZ3R2Y7JD7GOAKTQ:-NHkzJdr7QWa4T1dy28n:b-66jbou) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L11)):*

I don't see the need for any "types" of streams. The CAN table can just subscribe to the "CAN" stream as-is (since there is only the one), and generally in the future things should only subscribe to the explicit streams they are designed to subscribe to. This currently isn't really useful for "dynamic" displays anyway since it just gets called once at startup before we have any idea what streams there actually will be.
</blockquote></details>

I do agree with the idea that we need not to have types as a field in streams. However, the way we deal with CAN stream as we do now (unpack the payload at the plot end) in my opinion isn't really the way we should move forward, since then parser didn't really "parse" anything for CAN, rather it just sends over the package of payload wholesale. But the rewrite of that is another PR of itself, so there's nothing else to add to here other than "we are using 'channel_name|stream_name'" as a convention now (where CAN is just CAN).

PCModeActivate (commented):
Comments I left unresolved are mostly on stuff not written by me, and so I'd rather let @darth-raazi address them or put in his two cents before we come to a conclusion. (Of course, if he doesn't do anything, I'll have to do it either way.)

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NHl2iHR6tSo31opqr60:b-hilmt6)* status: 5 of 12 files reviewed, 10 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`sinks/dashboard/parsers.py` line 22 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbvjupBfaPBUzuI1uJ:-NHl3JAe475aQ7LPDh9n:bqnsdht) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L22)):*

Re: the issue with DAQ timestamps, I think it's possible to make a class and decorate one of its static methods so you can use class level variables to store state such as the first time received (like the old parser did). Probably worth investigating and seeing how complex the code ends up as, since storing state is useful.
</blockquote></details>

This looks more like a "feature" discussion (aka, yet another PR), but yes, if we have a need to store DAQ states then it would be quite helpful. Additionally, if we are to rewrite this alongside changing the way parse does things, we should probably also deal with the way CAN is parsed, which I think is worthy of further discussion.

BluCodeGH (changes requested):
> Reviewed 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: all files reviewed, 9 unresolved discussions (waiting on @Darth-Raazi, @KavinSatheeskumar, and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 22 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbvjupBfaPBUzuI1uJ:-NHk1MDKAduItpEKc8qn:b-dxcbts) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L22)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

This looks more like a "feature" discussion (aka, yet another PR), but yes, if we have a need to store DAQ states then it would be quite helpful. Additionally, if we are to rewrite this alongside changing the way parse does things, we should probably also deal with the way CAN is parsed, which I think is worthy of further discussion.
</blockquote></details>

Makes sense.
___
*[`sinks/dashboard/parsers.py` line 31 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaMYWX8FGzBhmV97MH:-NHk1ZaO0u9jKokdF02u:b-7sksa5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L31)):*

On the contrary it should be just after `register`, before any of the parsers. Don't want to have to scroll through all of that to see the common logic.
</blockquote></details>

bump.
___
*[`sinks/dashboard/items/can_message_table.py` line 238 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgOVz3-zd7_JkMMkMz:-NHgOVz3-zd7_JkMMkN-:b-bb4obr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/can_message_table.py#L238)):*
> ```Python
>             self.payloadQ.pop(0)
> 
>         message = self.get_msg()
> ```

??? You append and then immediately use the thing you just appended? And `payloadQ` isn't used elsewhere? Wouldn't `message = canSeries[1]` be equivalent?



_Code quote:_
```Python
        self.payloadQ.append(canSeries)

        if (len(self.payloadQ) > 50):
            self.payloadQ.pop(0)

        message = self.get_msg()
```
___
*[`sinks/dashboard/items/subscriber.py` line 4 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNX7vC6Zi70BKMILk:-NHk2wOK107Qn2127MXb:baicc76) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L4)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

I don't think we should be feeding back the `on_data_update` as a callback, since practically every "subscriber" "subclass" (airquotes because subscriber is gone now) should have its own `on_data_update`. It will be called upon update, so it doesn't need to be fed back into the publisher. Subscriber used to be an explicit inheritance structure that was designed to allow an observer to listen in on more than one stream, other than that it just calls `on_data_update`. In any case the observer classes should always have an  `on_data_update`, so it should be fine to leave it to call `on_data_update` from publisher directly.
</blockquote></details>

Passing a callback is a simpler and more extensible way of doing things. I don't see how needing an `on_data_update` is relevant, a callback is required in both cases and this way gives the flexibility to name it whatever / have multiple.
___
*[`sinks/dashboard/publisher.py` line 11 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHaRZ3R2Y7JD7GOAKTQ:-NHbxfD28m_ZKnbxLxr4:b-li58ft) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/publisher.py#L11)):*
<details><summary><i>Previously, PCModeActivate (Robert Cai) wrote…</i></summary><blockquote>

I do agree with the idea that we need not to have types as a field in streams. However, the way we deal with CAN stream as we do now (unpack the payload at the plot end) in my opinion isn't really the way we should move forward, since then parser didn't really "parse" anything for CAN, rather it just sends over the package of payload wholesale. But the rewrite of that is another PR of itself, so there's nothing else to add to here other than "we are using 'channel_name|stream_name'" as a convention now (where CAN is just CAN).
</blockquote></details>

Yeah I agree that the way we are doing CAN is wrong, but it doesn't require this `_type` logic currently since there is only one CAN stream. So  imo better to remove the unnecessary logic here so the future architecture is easier to build on.
___
*[`sinks/dashboard/publisher.py` line 9 at r7](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHk7CZlC999wPtKM0iA:-NHk7CZlC999wPtKM0iB:b-nptl4f) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/9d633ba627bdf0038b29f5db9bb366f9b3629119/sinks/dashboard/publisher.py#L9)):*
> ```Python
>         self.streams = {}
> 
>     def get_all_stream(self, _type=""):
> ```

Correct pluralization would be `get_all_streams`.

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: all files reviewed, 11 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 14 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHbv5-z--flst2P5ZTv:-NHoIdhD8aSJYbMoWxgp:bel9g7d) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/657e415ffffb78db6dcbf2c21d3e0a4f0a210c80/sinks/dashboard/parsers.py#L14)):*

I think this was mentioned in the other PR but (unlike with parsley where this code came from) we actually want the ability to have multiple parsers for the same message channel. I'd make `_func_map` a dictionary mapping channels to lists of parser functions.
</blockquote></details>

Very fair, 

One thing I will mention right now is a plan we have in the works for something I like to call "parser-deferal"


Basically, the CAN parser needs to have a whole bunch of different parsers for different CAN messages, that all get sent over the CAN channel. ERGO, we are thinking of having the can_parse function recursively call parse with a different channel. SO that we we can add this functionality whist adding a minimal amount of boiler plate
___
*[`sinks/dashboard/items/can_message_table.py` line 179 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgLg_ZAHY3YLQshhmU:-NHgLg_ZAHY3YLQshhmV:b-qmxtrw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/can_message_table.py#L179)):*
> ```Python
>         self.props = props
> 
>         self.payloadQ = []
> ```

Yo do not need a queue, but this is not something we should do in this PR
___
*[`sinks/dashboard/items/can_message_table.py` line 231 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgM1lx2e0VRUe6c6C3:-NHgM1lx2e0VRUe6c6C4:bgbh1ki) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/can_message_table.py#L231)):*
> ```Python
>     def on_data_update(self, canSeries):
>         # adding this line because canSeries is currently in the form of [timestamp, payload], and we want only the payload
>         canSeries = canSeries[1]
> ```

Kinda sigma, just make the LHS a different variable name
___
*[`sinks/dashboard/items/subscriber.py` line 4 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNX7vC6Zi70BKMILk:-NHoJXzc61sK2QyTK65N:b-4o50hh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L4)):*

Passing a callback is a simpler and more extensible way of doing things. I don't see how needing an `on_data_update` is relevant, a callback is required in both cases and this way gives the flexibility to name it whatever / have multiple.
</blockquote></details>

How do you plan to handle unsubscribing? the reason I switched away from a callback to explict subscriber classes is because it makes unsubscribe easier. If we are allowing subscriber to subscribe to multiple different things, I'd forsee some additional business end logic which could be hidden behind an abstract subscribe class.

Note, we need an unsubscribe since that needs to happen when dashitems are removed from the display

KavinSatheeskumar (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: all files reviewed, 11 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @PCModeActivate)

___
*[`sinks/dashboard/items/subscriber.py` line 4 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgNX7vC6Zi70BKMILk:-NHoKqc99GU8-iAtRF2g:b-b7x2md) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/subscriber.py#L4)):*

How do you plan to handle unsubscribing? the reason I switched away from a callback to explict subscriber classes is because it makes unsubscribe easier. If we are allowing subscriber to subscribe to multiple different things, I'd forsee some additional business end logic which could be hidden behind an abstract subscribe class.

Note, we need an unsubscribe since that needs to happen when dashitems are removed from the display
</blockquote></details>

This is just a personal opinion. But I find the idea of passing functions as arguments in an object oriented language kind of bad. As a result, I tend to favour passing objects with member functions rather than functions

Darth-Raazi:
Sorry, I didn't get any notifications from this PR and forgot to check cause I with busy with school. @PCModeActivate I noticed you were talking about leaving some things unresolved for me to fix. What exactly do you need me to address?

BluCodeGH (changes requested):
> Reviewed 7 of 7 files at r8, all commit messages.
> Dismissed @KavinSatheeskumar from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71)* status: all files reviewed, 9 unresolved discussions (waiting on @Darth-Raazi, @KavinSatheeskumar, and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 40 at r8](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NIFx6guBlR-yx-LEApX:-NIFx6guBlR-yx-LEApY:b-oxeur1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/584d51c2245033153074715b1cd7271684814474/sinks/dashboard/parsers.py#L40)):*
> ```Python
>             if msg_channel in Register.func_map:
>                 Register.func_map[msg_channel].append(func)
>             Register.func_map[msg_channel] = [func]
> ```

Currenly this overrides the func map every time lol. Imo its cleaner to check if the channel is _not_ in `func_map` and if so add it as an empty list, then append either way.
___
*[`sinks/dashboard/parsers.py` line 49 at r8](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NIFxVIe07KiQbW4U6ow:-NIFxVIe07KiQbW4U6ox:bnirrtq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/584d51c2245033153074715b1cd7271684814474/sinks/dashboard/parsers.py#L49)):*
> ```Python
>                 # For an explanation, please refer to later block of comments
>                 stream_message_pair_list = func(msg_payload)
>                 for stream_name, timestamp, parsed_message in stream_message_pair_list:
> ```

`for stream_name, timestamp, parsed_message in func(msg_payload)` is easier to read imo.
___
*[`sinks/dashboard/items/can_message_table.py` line 179 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NHgLg_ZAHY3YLQshhmU:-NIFz1aT1s5e-5_yFzqq:b1t92j2) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4a62bc544a6289736b78fe89e7b18bb5d8467037/sinks/dashboard/items/can_message_table.py#L179)):*

Yo do not need a queue, but this is not something we should do in this PR
</blockquote></details>

This is now unused and should be removed.
___
*[`sinks/dashboard/publisher.py` line 14 at r8](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NIFy6338YxBsj_EaTDA:-NIFy6338YxBsj_EaTDB:bk9hnhe) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/584d51c2245033153074715b1cd7271684814474/sinks/dashboard/publisher.py#L14)):*
> ```Python
>         self.streams = {}
> 
>     def get_all_streams(self, _type=""):
> ```

unused argument
___
*[`sinks/dashboard/publisher.py` line 19 at r8](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-NIFyMbn1qJb3Lu-cjCk:-NIFyMbn1qJb3Lu-cjCl:bdz1h57) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/584d51c2245033153074715b1cd7271684814474/sinks/dashboard/publisher.py#L19)):*
> ```Python
>     def subscribe(self, stream, callback):
>         if stream in self.streams:
>             self.streams[stream].append(callback)
> ```

We should be able to subscribe to a stream that hasn't been published yet. The previous logic was fine:
```
        if stream not in self.streams:
            self.streams[stream] = []
        self.streams[stream].append(observer)
```

BluCodeGH (approved):
:lgtm:

> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NIPkGlx09TCUuDoOF6h:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Darth-Raazi)

Darth-Raazi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NIUNY_zEHF7BhTRAaLx:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @PCModeActivate)

KavinSatheeskumar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/71#-:-NIVRkykCvgK0-cAreTv:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @PCModeActivate)

PR #72, opened 2022-11-13
Fix Parsers

Darth-Raazi:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -10,13 +10,12 @@ from pyqtgraph.dockarea.DockArea import DockArea
 from pyqtgraph.Qt import QtWidgets
 from pyqtgraph.Qt.QtGui import QVBoxLayout, QMenuBar
 
-#import parsers
 from items.plot_dash_item import PlotDashItem
 from items.can_message_table import CanMsgTableDashItem
 from omnibus.util import TickCounter
 from utils import prompt_user
 
-# "Temorary Global Constant"
+# "Temporary Global Constant"
 
 item_types = [
     PlotDashItem,
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -10,7 +10,7 @@ from pyqtgraph.dockarea.DockArea import DockArea
 from pyqtgraph.Qt import QtWidgets
 from pyqtgraph.Qt.QtGui import QVBoxLayout, QMenuBar
 
-from parsers import Parser
+#import parsers
 from items.plot_dash_item import PlotDashItem
 from items.can_message_table import CanMsgTableDashItem
 from omnibus.util import TickCounter
--- a/sinks/dashboard/items/dashboard_item.py
+++ b/sinks/dashboard/items/dashboard_item.py
@@ -20,7 +20,8 @@ class DashboardItem(QtWidgets.QWidget):
         '''
         Return a nicer name for the Dash Item instead of the class name
         '''
-        raise NotImplementedError
+        #raise NotImplementedError
+        return "Kavin was here"
 
     def get_props(self) -> typing.Any:
         """
--- a/sinks/dashboard/items/dashboard_item.py
+++ b/sinks/dashboard/items/dashboard_item.py
@@ -20,8 +20,7 @@ class DashboardItem(QtWidgets.QWidget):
         '''
         Return a nicer name for the Dash Item instead of the class name
         '''
-        #raise NotImplementedError
-        return "Kavin was here"
+        raise NotImplementedError
 
     def get_props(self) -> typing.Any:
         """
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -28,7 +28,7 @@ class PlotDashItem (DashboardItem):
         # if no properties are passed in
         # prompt the user for them
         if self.props == None:
-            items = [sensor for sensor, series in temp_series_dict.items()]
+            items = list(temp_series_dict.keys())
 
             self.props = prompt_user(
                 self,
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -28,16 +28,15 @@ class PlotDashItem (DashboardItem):
         # if no properties are passed in
         # prompt the user for them
         if self.props == None:
-            items = [f"{channel}" for channel, series in temp_series_dict.items()]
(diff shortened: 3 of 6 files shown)


BluCodeGH (changes requested):
Yeah so apart from some style comments the main idea here looks good, but it's too hacky with this `temp_series_dict` stuff to merge on its own. Let's combine it with the series rewrite to do one big ~happy family~ refactor of the whole system.

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-:-NGoJWPY29hwCLBHoUAp:b-ooyvr9)* status: all files reviewed, 7 unresolved discussions (waiting on @Darth-Raazi, @KavinSatheeskumar, and @PCModeActivate)

___
*[`sinks/dashboard/dashboard.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoHZzn0EG2adq0Gur0:-NGoHZzn0EG2adq0Gur1:b-u8ccuv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/dashboard.py#L13)):*
> ```Python
> from pyqtgraph.Qt.QtGui import QVBoxLayout, QMenuBar
> 
> #import parsers
> ```

no outdated comments pls
___
*[`sinks/dashboard/parsers.py` line 17 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoI_Y7D2eJBj9sWYAW:-NGoI_Y7D2eJBj9sWYAX:b-2iv2vr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/parsers.py#L17)):*
> ```Python
>         return self[key]
> 
> temp_series_dict = SeriesDefaultDict()
> ```

Yeah this isn't gonna fly lol, let's combine this with the series changes as one big PR.
___
*[`sinks/dashboard/parsers.py` line 19 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoImV60bu-fGILMeQo:-NGoImV60bu-fGILMeQp:bbmxch8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/parsers.py#L19)):*
> ```Python
> temp_series_dict = SeriesDefaultDict()
> 
> _func_map = {}
> ```

This should be declared right above `register()` so its clear where it's used.
___
*[`sinks/dashboard/parsers.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoITfD5Sr9dsIPj-Th:-NGoITfD5Sr9dsIPj-Ti:bhyw5w8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/parsers.py#L40)):*
> ```Python
> @register("DAQ")
> def daq_parser(msg_data):
>     global start
> ```

It's probably preferable to just not adjust DAQ timestamps than use `global`.
___
*[`sinks/dashboard/parsers.py` line 58 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoIwlP2cfhb-tSDD7K:-NGoIwlP2cfhb-tSDD7L:bn1wdy0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/parsers.py#L58)):*
> ```Python
> 
> 
> ####################################### OLD CODE #############################################
> ```

Kill it with fire
___
*[`sinks/dashboard/items/dashboard_item.py` line 24 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoJ8k44T3yhIswEoMr:-NGoJ8k44T3yhIswEoMs:b-1oguok) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/items/dashboard_item.py#L24)):*
> ```Python
>         '''
>         #raise NotImplementedError
>         return "Kavin was here"
> ```

no :slightly_smiling_face: 
___
*[`sinks/dashboard/items/plot_dash_item.py` line 31 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-NGoJIRBB7tOsyW5mbML:-NGoJIRBB7tOsyW5mbMM:b-g5t77e) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dd4d909fa7dc77fadf2e8503acb31e9f8ac41288/sinks/dashboard/items/plot_dash_item.py#L31)):*
> ```Python
>         # prompt the user for them
>         if self.props == None:
>             items = [sensor for sensor, series in temp_series_dict.items()]
> ```

consider calling `list(temp_series_dict.keys())` instead of this.

Darth-Raazi:
What's the problem with using `global`?

Also I'm leaving part of the old code in `parsers.py` for testing purposes, dashboard doesn't run without it cause of `CanDisplayParser`. It'll be deleted once `Parsley` is made into a big library.

Also, this PR will be deleted very soon and changes will instead be put on `dash-data-processing`.

BluCodeGH (commented):
`global` is bad because it means your program has mutable global state, ie global state that can change over the runtime of the program. Take a look at the accepted answer here (with the context that it is talking about _mutable_ global state) for more info: https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil.

Now, that might leave you wondering why `_func_map` is acceptable. The answer is that it's pretty borderline, if there was an easy way to wrap its state up we should (honestly just googled looks like it isn't too hard to use a class as a decorator). The reason it's survived this long is that `_func_map` is only changed during program initialization. The code inside the decorator function that registers the function into the map actually runs _before *any*_ of the rest of the code in omnibus, it runs as python is parsing the file (or maybe just after, idk the details). So this means that once the main code actually starts running `_func_map` is fully set up and never changes.

Lastly for future reference you can click on the big "Reviewable" button to reply to my comments individually.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/72#-:-NGtLLstBlb6AKjttR23:b-mgclmn)* status: all files reviewed, 2 unresolved discussions (waiting on @Darth-Raazi, @KavinSatheeskumar, and @PCModeActivate)

PR #73, opened 2022-11-15
Fix Parsers Again Cause Kavin Said

Darth-Raazi:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -10,13 +10,12 @@ from pyqtgraph.dockarea.DockArea import DockArea
 from pyqtgraph.Qt import QtWidgets
 from pyqtgraph.Qt.QtGui import QVBoxLayout, QMenuBar
 
-#import parsers
 from items.plot_dash_item import PlotDashItem
 from items.can_message_table import CanMsgTableDashItem
 from omnibus.util import TickCounter
 from utils import prompt_user
 
-# "Temorary Global Constant"
+# "Temporary Global Constant"
 
 item_types = [
     PlotDashItem,
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -10,7 +10,7 @@ from pyqtgraph.dockarea.DockArea import DockArea
 from pyqtgraph.Qt import QtWidgets
 from pyqtgraph.Qt.QtGui import QVBoxLayout, QMenuBar
 
-from parsers import Parser
+#import parsers
 from items.plot_dash_item import PlotDashItem
 from items.can_message_table import CanMsgTableDashItem
 from omnibus.util import TickCounter
--- a/sinks/dashboard/items/dashboard_item.py
+++ b/sinks/dashboard/items/dashboard_item.py
@@ -20,7 +20,8 @@ class DashboardItem(QtWidgets.QWidget):
         '''
         Return a nicer name for the Dash Item instead of the class name
         '''
-        raise NotImplementedError
+        #raise NotImplementedError
+        return "Kavin was here"
 
     def get_props(self) -> typing.Any:
         """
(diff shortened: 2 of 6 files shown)


KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @PCModeActivate)

___
*[`sinks/dashboard/parsers.py` line 28 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73#-NH6pDCs5rEVZxwIitKo:-NH6pDCs5rEVZxwIitKp:b-pb3ezz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f7164d3de78fd6620611507fc1f50d6dc411384c/sinks/dashboard/parsers.py#L28)):*
> ```Python
>         for msg_channel in msg_channels:
>             for func in _func_map:
>                 if func.startswith(msg_channel):
> ```

I am not the biggest fan of this behavior. We kinda intend to be able to send messages both over channels "A" and "AB" and have things respond accordingly.

I might change it like this, only throw an exception if msg_channel is a key in _func_map, otherwise just add it.

This way, the message gets passed to all parsers who's corresponding channel is a prefix of the message.

The only code change this would require is replacing this for loop with an if statement



_Code snippet:_
```Python
if msg_channel in _func_map:
    raise KeyError(...)
```
___
*[`sinks/dashboard/parsers.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73#-NH6rBAu-HYoBscqSIqs:-NH6rBAv-v4aLxO9D12e:b-pvhgtd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f7164d3de78fd6620611507fc1f50d6dc411384c/sinks/dashboard/parsers.py#L40)):*
> ```Python
> 
>     for sensor, data in msg_data["data"].items():
>         temp_series_dict[sensor].add(timestamp, sum(data)/len(data))
> ```

I am thinking of adding this change for testability purposes.

Maybe we should have the daq_parse function just return a list of (series-name, timestamp, parsed-message) tripples.

then the parse function
* Calls this function
* grabs this return value
* puts each parsed-message in the corresponding series with the correct timestamp

This will make writing unit tests easier
___
*[`sinks/dashboard/parsers.py` line 46 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73#-NH6rgE888uBkCyPZKT1:-NH6rgE888uBkCyPZKT2:b-1tog0i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f7164d3de78fd6620611507fc1f50d6dc411384c/sinks/dashboard/parsers.py#L46)):*
> ```Python
>     for func in _func_map:
>         if msg_channel.startswith(func):
>             _func_map[func](msg_payload)
> ```

Inline with my previous comment, the way this will work is like this



_Code snippet:_
```Python
series_message_pair_list = _func_map[func](msg_payload)
for series_name, timestamp, parsed_message in series_message_pair_list:
    temp_series_dict[series_name].add(timestamp, parsed_message)
```
___
*[`sinks/dashboard/parsers_test.py` line 29 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/73#-NH6sI1y4fIV0S2OWWAZ:-NH6sI1y4fIV0S2OWWA_:b-t8p8xn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f7164d3de78fd6620611507fc1f50d6dc411384c/sinks/dashboard/parsers_test.py#L29)):*
> ```Python
>             "data": {"SENSOR 1": [1, 2, 6], "SENSOR 2": [4, 5, 6]}
>         }
>         parsers.parse("DAQ",payload)
> ```

The comments above were thought up in response to this

BluCodeGH:
Closing as superceded by #71.

PR #74, opened 2022-11-15
Update launcher to support macOS

Darth-Raazi:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74)

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -1,37 +1,59 @@
-import sys
-from subprocess import Popen, CREATE_NEW_CONSOLE
+from subprocess import Popen
+import platform
 
+# Set the correst source based on whether or not the user is testing Omnibus
+test = ''
+while test != 'y' and test != 'n':
+    test = input("Is this a test of Omnibus? [y/n] ").lower()
 
-profiles = {
-    "test": ['python -m omnibus',
-             'python sinks/plot/main.py',
-             'python sources/fakeni/main.py'],
-    "texas": ['python sources/parsley/main.py $arg',
-              'python sources/ni/main.py',
-              'python -m omnibus']
-    #   Add other profiles in here
-    #   If any commands require arguments to be passed in, write $arg in its place
+if (test == 'y'):
+    source = 'fakeni'
+else:
+    source = 'ni'
 
-}
+# Darwin is macOS, Windows is Windows, Linux... maybe later
+if platform.system() == "Darwin":
+    import os
+    import time
 
+    # Get the current directory (where omnibus is located)
+    omnibus_path = os.getcwd()
 
-try:
-    if sys.argv[1] == "_wrap":
-        p = Popen(sys.argv[2])
-        if p.wait() != 0:
-            input('Press enter to quit')
+    # Change this to if you named your virtual environment something else
+    venv_name = 'venv'
 
-    else:
-        selection = sys.argv[1]
-        processes = profiles[selection]
-        args = sys.argv[2:]
+    # Change this if you want to use a different terminal on mac
+    terminal = 'terminal'
 
-        for process in processes:
-            while '$arg' in process:
-                process = process.replace('$arg', args[0], 1)
-                args.pop(0)
-            Popen(f'python launcher.py _wrap "{process}"', creationflags=CREATE_NEW_CONSOLE)
+    # Writes correct paths to scripts
+    with open("omnibus.sh", 'w') as f:
+        f.write(f'cd {omnibus_path}\nsource {venv_name}/bin/activate\npython -m omnibus\ntrap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT')
 
-except (KeyError, IndexError):
-    print('Please enter a single valid profile selection (ex. python launcher.py texas)')
-    print('Ensure that you have entered the correct amount of args required for each script in the right order')
+    with open("source.sh", 'w') as f:
+        f.write(f'cd {omnibus_path}\nsource {venv_name}/bin/activate\npython sources/{source}/main.py\ntrap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT')
+
+    with open("dashboard.sh", 'w') as f:
+        f.write(f'cd {omnibus_path}\nsource {venv_name}/bin/activate\npython sinks/dashboard/main.py\ntrap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT')
+
+    # Open new terminal windows and run the scripts that run python commands
+    Popen(['open', '-a', terminal, 'omnibus.sh']).communicate()
+    Popen(['open', '-a', terminal, 'source.sh']).communicate()
+    
+    # Dashboard needs to be delayed slightly so omnibus and sink can find each other
+    time.sleep(0.5)
+    Popen(['open', '-a', terminal, 'dashboard.sh']).communicate()
+
+elif platform.system() == "Windows":
+    # NOT TESTED YET - need someone with a Windows machine
+    from subprocess import CREATE_NEW_CONSOLE
+
+    commands = ['python -m omnibus',
+                f'python sources/{source}/main.py',
+                'python sinks/dashboard/main.py']
+
+    for command in commands:
+        Popen(command, creationflags=CREATE_NEW_CONSOLE)
+
+else:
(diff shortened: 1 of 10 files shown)


BluCodeGH (changes requested):
Good thought with the formatting run but this will cause merge conflicts when any in-progress PRs touch any of these files. Dealing with reformatting everything in a way that won't cause conflicts is a bigger task and shouldn't be included in this PR.

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-:-NGyZCCTAoNgeymXxM4L:b53ohfc)* status: all files reviewed, 2 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`launcher.py` line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYGMU4Y73XEE9wJAb:-NGyYGMU4Y73XEE9wJAc:b-m5robc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L12)):*
> ```Python
>     source = 'fakeni'
> else:
>     source = 'ni'
> ```

The profiles mechanism is more extensible than this hardcoded prompt, why did you remove it?
___
*[`launcher.py` line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYOBrDoyXNIaqFzHG:-NGyYOBrDoyXNIaqFzHH:bx3pt3e) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L15)):*
> ```Python
> 
> # Darwin is macOS, Windows is Windows, Linux... maybe later
> if platform.system() == "Darwin":
> ```

Having entirely different code for different platforms is not good practice. Also this OSX mechanism with writing to sh files and executing is pretty damn hacky (depends on the venv name, terminal executable, etc), I'm not a fan of merging this approach. We should find a better cross-platform way to do this or change how the launcher works (eg it doesn't necessarily need to open new windows, it just needs to run the subprocesses, show errors from any of them and let them exit together).

Darth-Raazi (commented):
Ah darn. Kavin said it wouldn't cause any merge issues, but I can revert the changes for the next commit. I didn't realize the formatting script ran for the entire repo instead of just the files I was working on.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-:-NGyapfxCV1L-7ryVnil:b-2ytr44)* status: all files reviewed, 2 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`launcher.py` line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYGMU4Y73XEE9wJAb:-NGybEl88eg6C7AQwQAH:bu8wane) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L12)):*

The profiles mechanism is more extensible than this hardcoded prompt, why did you remove it?
</blockquote></details>

The way the profiles mechanism was set works well on Windows but breaks on macOS because of terminal shenanigans. Calling the same file over and over again with different arguments was causing problems. The only thing that was really changing was which source we were using, so I opted for this approach. This also allows for the user to not need to know how `launcher.py` works (command-line arguments needed) and can simply run the file and respond to the prompts. If we need more profiles, perhaps the user can just be prompted more?
___
*[`launcher.py` line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYOBrDoyXNIaqFzHG:-NGydZHJ1Zy2cARdXvZJ:b-z4kiim) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L15)):*

Having entirely different code for different platforms is not good practice. Also this OSX mechanism with writing to sh files and executing is pretty damn hacky (depends on the venv name, terminal executable, etc), I'm not a fan of merging this approach. We should find a better cross-platform way to do this or change how the launcher works (eg it doesn't necessarily need to open new windows, it just needs to run the subprocesses, show errors from any of them and let them exit together).
</blockquote></details>

Here's the thing: I've looked into this a lot and there doesn't seem to be a cross-platform solution because of how the different OS's handle code execution and shells. The `CREATE_NEW_CONSOLE` flag is amazing and does exactly what we want, but it's only supported on Windows. I tried having the files run concurrently in the background but they don't play nice with each other. Unless there's a solution that I'm unaware of, I think having everything run together would require changing a lot of how Omnibus sends and receives data under the hood. If you know something I don't, please share it cause it would be very helpful lol. 

I mostly wanted `launcher.py` to work for me because I ran Omnibus probably 100 times while rewriting `parsers.py` and it was really annoying to have to start 3 different files every single time. This works well for me so I thought it could be helpful for anyone else running macOS as well, but if you think we should leave it like it was before, I can close the PR and keep the changes just on my local machine for myself.

BluCodeGH (commented):
Yeah the formatting script typically should only edit the files your PR changed, but I think the funky dashboard branching stuff messes with that logic.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-:-NGyhv0jEhFW6OEm9EkZ:b-63xgmx)* status: all files reviewed, 2 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`launcher.py` line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYGMU4Y73XEE9wJAb:-NGyiAIX4M5LKLVNzEQO:b8vtkhs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L12)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

The way the profiles mechanism was set works well on Windows but breaks on macOS because of terminal shenanigans. Calling the same file over and over again with different arguments was causing problems. The only thing that was really changing was which source we were using, so I opted for this approach. This also allows for the user to not need to know how `launcher.py` works (command-line arguments needed) and can simply run the file and respond to the prompts. If we need more profiles, perhaps the user can just be prompted more?
</blockquote></details>

The profiles mechanism is different from the _wrap functionality, the only reason _wrap exists is to keep the terminal windows open if it crashes so we can see the error message / traceback. Adding more manual prompting and hardcoding is not a good extensible solution, I think the best way to do it is to take a CLI arg for the profile to use and if the argument isn't given then prompt the user and ask them to choose a profile.
___
*[`launcher.py` line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/74#-NGyYOBrDoyXNIaqFzHG:-NGygNg_2bqN1EvBBagE:b-fddxxz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f3ed51810c9a8147697f298503faf611203a0cdd/launcher.py#L15)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

Here's the thing: I've looked into this a lot and there doesn't seem to be a cross-platform solution because of how the different OS's handle code execution and shells. The `CREATE_NEW_CONSOLE` flag is amazing and does exactly what we want, but it's only supported on Windows. I tried having the files run concurrently in the background but they don't play nice with each other. Unless there's a solution that I'm unaware of, I think having everything run together would require changing a lot of how Omnibus sends and receives data under the hood. If you know something I don't, please share it cause it would be very helpful lol. 

I mostly wanted `launcher.py` to work for me because I ran Omnibus probably 100 times while rewriting `parsers.py` and it was really annoying to have to start 3 different files every single time. This works well for me so I thought it could be helpful for anyone else running macOS as well, but if you think we should leave it like it was before, I can close the PR and keep the changes just on my local machine for myself.
</blockquote></details>

I think the way to upgrade the launcher is to control the different processes ourselves via popen. We can popen the desired subprocess, redirect only standard error to the console, and automatically kill the other two when one exists or when the user presses ctrl+c.

Also top tip, you can leave all of omnibus running except for the part of it you're developing. Like for example run the fakeni source and omnibus server and minimize those terminals, and then just start and restart the dashboard, like any other program.

PR #75, opened 2022-11-17
Change color when hits limit

yxyyeah:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -213,17 +213,16 @@ class Dashboard(QtWidgets.QWidget):
     def switch(self):
         self.save()
         filename = prompt_user(
-                self,
-                "New File Name",
-                "Enter the name of the file which you wish to load",
-                "items",
-                self.filename_cache,
-                True
-            )
+            self,
+            "New File Name",
+            "Enter the name of the file which you wish to load",
+            "items",
+            self.filename_cache,
+            True
+        )
 
         if filename == None:
             return
-
 
         # If the filename entered is not valid
         # this exhibits the behaviour of creating
@@ -240,7 +239,7 @@ class Dashboard(QtWidgets.QWidget):
         self.counter.tick()
 
         # Filter to 5 frames per update on analytics
-        if not(self.counter.tick_count() % 5):
+        if not (self.counter.tick_count() % 5):
             fps = self.counter.tick_rate()
 
         self.callback()
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -40,19 +40,37 @@ class PlotDashItem (DashboardItem):
                 "Data Series",
                 "The series you wish to plot",
                 "items",
-                items
-                )
+                items,
+            )
+            if not channel_and_series:
+                raise Exception
+            
+            threshold_input = prompt_user(
+                self,
+                "Threshold Value",
+                "Set an upper limit",
+                "number"
+            )
+            if threshold_input == None:
+                raise Exception
+
             self.props = channel_and_series.split("|")
+            self.props.append(threshold_input)
 
         # subscribe to series dictated by properties
         self.series = Parser.get_series(self.props[0], self.props[1])
         self.subscribe_to_series(self.series)
 
+        # set the limit for triggering red tint area
+        self.limit = self.props[2]
+
         # create the plot
         self.plot = pg.PlotItem(title=self.series.name, left="Data", bottom="Seconds")
         self.plot.setMouseEnabled(x=False, y=False)
         self.plot.hideButtons()
+        # create data curve and warning line
         self.curve = self.plot.plot(self.series.times, self.series.points, pen='y')
+        self.warning_line = self.plot.plot([], [], fillLevel=self.limit*10, brush=(255,0,0,50),pen='r')
 
         # create the plot widget
         self.widget = pg.PlotWidget(plotItem=self.plot)
@@ -68,19 +86,27 @@ class PlotDashItem (DashboardItem):
         min_time = t - config.GRAPH_DURATION + config.GRAPH_STEP
         max_time = t + config.GRAPH_STEP
 
-        #filter the times
-
+        # filter the times, points
         times = [series.times[i] for i in range(series.size) if (
             series.times[i] >= min_time and series.times[i] <= max_time)]
-
         points = [series.points[i] for i in range(series.size) if (
             series.times[i] >= min_time and series.times[i] <= max_time)]
+        
+        min_point = min(points)
+        max_point = max(points)
+        
+        #set the displayed range of Y axis
+        self.plot.setYRange(min_point,max_point,padding=0.1)
 
-        self.curve.setData(times, points)
+        # plot the warning line
+        self.warning_line.setData(times,[self.limit] * len(points))
+            
+        # plot the data curve
+        self.curve.setData(times,points)
 
         # current value readout in the title
         self.plot.setTitle(
-            f"[{sum(points)/len(points): <4.4f}] [{series.points[-1]}] {series.name} {series.desc and (series.desc + ' ') or ''}")
+            f"[{sum(points)/len(points): <4.4f}] [{points[-1]}] {series.name} {series.desc and (series.desc + ' ') or ''}")
 
     def get_props(self):
         return self.props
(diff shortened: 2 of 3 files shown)


BluCodeGH (changes requested):
Looks great! A few nitpicks:

Looks like you still have some trailing whitespace on blank lines, maybe run autopep8 manually if the format script isn't working? Eg:


> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-:-NH2yapP9txS5koFUijp:b-61dq6q)* status: all files reviewed, 2 unresolved discussions (waiting on @yxyyeah)

___
*[`sinks/dashboard/items/plot_dash_item.py` line 46 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2xpp5EgrJXHqKhf5j:-NH2xpp5EgrJXHqKhf5k:b8k1ehj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L46)):*
> ```Python
>             )
>             if not channel_and_series:
>                 raise Exception
> ```

Under what circumstances could this not give a value? Seems like that is a bug to fix in prompt_user such that it ensures we get a valid return. 
___
*[`sinks/dashboard/items/plot_dash_item.py` line 102 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2yL_y-U-tznV82Tw7:-NH2yL_y-U-tznV82Tw8:b-7a77tf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L102)):*
> ```Python
> 
>         # plot the warning line
>         self.warning_line.setData(times,[self.limit] * len(points))
> ```

What if we initialize the warning line with two points: (0, limit) and (verybignumber, limit)? Would it then always display and save the performance overhead of updating it every time?

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75)* status: all files reviewed, 3 unresolved discussions (waiting on @BluCodeGH and @yxyyeah)

___
*[`sinks/dashboard/items/plot_dash_item.py` line 46 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2xpp5EgrJXHqKhf5j:-NH6uT914lGevcnz9Knd:b-86zs87) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L46)):*

Under what circumstances could this not give a value? Seems like that is a bug to fix in prompt_user such that it ensures we get a valid return. 
</blockquote></details>

None is a sential signaling that the user quit their selection. The correct behaviour is not adding the dashitem to the screen. IE the user decides they don't wanna add the thing anymore.

The fact that we throw an exception instead of doing that is the real bug
___
*[`sinks/dashboard/items/plot_dash_item.py` line 55 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH6ukgL4Z6j-YHUCrcd:-NH6ukgM8iZCSGRgeM5-:bx9xufg) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L55)):*
> ```Python
>             )
>             if threshold_input == None:
>                 raise Exception
> ```

It should be possible for there not to be a threshold. 

Maybe we should prompt them for text instead, and if they input a non-number, then we default to having no threshold

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75)* status: all files reviewed, 4 unresolved discussions (waiting on @BluCodeGH and @yxyyeah)

___
*[`sinks/dashboard/items/plot_dash_item.py` line 52 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH6xtLoDiXQE36YxCpw:-NH6xtLoDiXQE36YxCpx:br6h6ty) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L52)):*
> ```Python
>                 "Threshold Value",
>                 "Set an upper limit",
>                 "number"
> ```

This is a bug on my code, but it seems that it only lets me select a multiple of 0.1, this would also be resolved by the below suggestion
___
*[`sinks/dashboard/items/plot_dash_item.py` line 102 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2yL_y-U-tznV82Tw7:-NH6ye9rDvuGTF5dMTk6:b-54kz20) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L102)):*

What if we initialize the warning line with two points: (0, limit) and (verybignumber, limit)? Would it then always display and save the performance overhead of updating it every time?
</blockquote></details>

This is a bug I found whilst poking at it.

If I set the limit to below the graph limits, it doesn't show. IDK if its worth fixing, but its worth pointing out

yxyyeah:
>

yxyyeah:
> What if we initialize the warning line with two points: (0, limit) and (verybignumber, limit)? Would it then always display and save the performance overhead of updating it every time?
Then the plot would always display the range from 0-verybignumber. In that case I may need to update the range of the x axis each time. Which one would perform better: updating the line or the axis? Also each time I can draw a line with 2 points (starting and finishing) instead of all points in times. Does this sounds better?

BluCodeGH:
Yeah i think two points sounds better. We should also try to fix the issue kavin mentioned with the fill disappearing when below the plot

BluCodeGH (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75)* status: all files reviewed, 4 unresolved discussions (waiting on @KavinSatheeskumar and @yxyyeah)

___
*[`sinks/dashboard/items/plot_dash_item.py` line 46 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2xpp5EgrJXHqKhf5j:-NHbsxyaBaz48tAeRI_B:b-q6gmpj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L46)):*

None is a sential signaling that the user quit their selection. The correct behaviour is not adding the dashitem to the screen. IE the user decides they don't wanna add the thing anymore.

The fact that we throw an exception instead of doing that is the real bug
</blockquote></details>

Is this something we should resolve here?
___
*[`sinks/dashboard/items/plot_dash_item.py` line 102 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/75#-NH2yL_y-U-tznV82Tw7:-NHbt-Vu29g3faEBtEix:b9zme7g) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ffa57bcefc3ec6c3baa6276eafcbccb5b3ce104/sinks/dashboard/items/plot_dash_item.py#L102)):*

This is a bug I found whilst poking at it.

If I set the limit to below the graph limits, it doesn't show. IDK if its worth fixing, but its worth pointing out
</blockquote></details>

Any idea how we can fix this? I think it's important to always have that red tint when the data is over the limit.

yxyyeah:
@BluCodeGH 

> Is this something we should resolve here?

New PR is on the way!

> Any idea how we can fix this? I think it's important to always have that red tint when the data is over the limit.

I think we do have the red tint when data is over the limit. So not an issue anymore

yxyyeah:
Changes of this PR has been merged into [prompt_for_properties PR](https://github.com/waterloo-rocketry/omnibus/pull/79). Maybe we can close this one

BluCodeGH:
Closing for tidiness, included in #79.

PR #81, opened 2022-12-03
Change savefile to JSON

Darth-Raazi:
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/81)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -2,6 +2,7 @@ import pickle
 import os
 import time
 import sys
+import json
 
 from pyqtgraph.Qt import QtCore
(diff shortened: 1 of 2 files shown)


BluCodeGH:
What's the motivation behind this PR? I'd like to avoid further reliance on the `item_types` "temporary global constant" if possible.

Darth-Raazi:
Kavin said to change the savefile type from binary to JSON so that it can be edited independently of the dashboard. That way we could preset how Omnibus would look before a CF or SF. I had to use the class conversion (using `item_types`) because the JSON library can't convert an `ObjectType` object. Maybe there's a better way to change `dashboard.py` so we don't have that global constant?

PR #86, opened 2022-12-13
Removed item_types

Darth-Raazi:
Removed global constant and added register function

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)


BluCodeGH (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)* status: all files reviewed, 4 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 19 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NJHlHIZ-FayiGuliU5O:-NJHlHIZ-FayiGuliU5P:bmv93oc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e5190cea94d5afe911cc757c79e987280a29cbd6/sinks/dashboard/dashboard.py#L19)):*
> ```Python
> 
> class Register:
>     item_map = {}
> ```

We only really need a list not a full map here.
___
*[`sinks/dashboard/dashboard.py` line 22 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NJHkjQ223sp22UQRJ2K:-NJHkjQ223sp22UQRJ2L:b-m83hxp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e5190cea94d5afe911cc757c79e987280a29cbd6/sinks/dashboard/dashboard.py#L22)):*
> ```Python
> 
>     def __init__(self, item_names):
>         if isinstance(item_names, str):
> ```

Since dashboarditems just need to be registered and added to a list we don't actually need to take any arguments here.
___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NJHkSup9xqCCvzV2iN3:-NJHkSuqDe5zfjM8cRz4:b-bni43o) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e5190cea94d5afe911cc757c79e987280a29cbd6/sinks/dashboard/dashboard.py#L36)):*
> ```Python
> @Register("PlotDashItem")
> def plot_dash_item():
>     return PlotDashItem
> ```

The `Register` class should go in its own file, so that the actual dashboard items can import it and register themselves (you can put a decorator on a class too). That removes the need for these (still hardcoded-ish) functions.
___
*[`sinks/dashboard/dashboard.py` line 173 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NJHlfa3757HXoI2cMUg:-NJHlfa3757HXoI2cMUh:b-z1kn7f) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e5190cea94d5afe911cc757c79e987280a29cbd6/sinks/dashboard/dashboard.py#L173)):*
> ```Python
>             # Convert the name of the class back into the class
>             # See save() method
>             for item_name, item_type in Register.item_map.items():
> ```

If we switch the `item_map` to a list we won't have the item names to use for saving/loading, you should be able to use `item.__name__` instead.

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)* status: 1 of 5 files reviewed, 6 unresolved discussions (waiting on @BluCodeGH and @Darth-Raazi)

___
*[`sinks/dashboard/dashboard.py` line 5 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKVo9XD7xBU1jtswfcU:-NKVo9XECHEzSVhW5Tsk:bh3vmga) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2df90f2f327535487508c43052001f2265a1d5e9/sinks/dashboard/dashboard.py#L5)):*
> ```Python
> import json
> 
> from items.register import Register
> ```

This is rather bitchy, but we already have another one of these register things in the parsers which is also called register. Simply for clarity I think the name had ought be changed to something else.

Somewhere, we will create documentation on this so called "Register Pattern" so that we don't have to call it register every time lol
___
*[`sinks/dashboard/items/register.py` line 2 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKVr--7CACEqf2rEAZl:-NKVr--7CACEqf2rEAZm:bgrky4s) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2df90f2f327535487508c43052001f2265a1d5e9/sinks/dashboard/items/register.py#L2)):*
> ```Python
> class Register:
>     item_list = []
> ```

part of me would prefer the item list to be a member of dashboard, rather than its own thing. This means we could use encapsulation and have Register be a member class of dashboard. Although, this is a subjective idea

BluCodeGH (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)* status: all files reviewed, 3 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 5 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKVo9XD7xBU1jtswfcU:-NKW0sTo88Hv_p_RdYgy:b-e83236) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2df90f2f327535487508c43052001f2265a1d5e9/sinks/dashboard/dashboard.py#L5)):*

This is rather bitchy, but we already have another one of these register things in the parsers which is also called register. Simply for clarity I think the name had ought be changed to something else.

Somewhere, we will create documentation on this so called "Register Pattern" so that we don't have to call it register every time lol
</blockquote></details>

I'd recommend renaming the file to `registry.py` and adding a method in it called `get_items()` that returns the list of items.
___
*[`sinks/dashboard/dashboard.py` line 148 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKW1b92A1hGPN6rGtiU:-NKW1b92A1hGPN6rGtiV:bvf3lw6) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2df90f2f327535487508c43052001f2265a1d5e9/sinks/dashboard/dashboard.py#L148)):*
> ```Python
>             for item_type in Register.item_list:
>                 if item["class"] == item_type.get_name():
>                     item["class"] = item_type
> ```

Cleaner / more understandable to call `self.add` followed by `break` within the if statement here rather than mutating `item`.
___
*[`sinks/dashboard/items/register.py` line 2 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKVr--7CACEqf2rEAZl:-NKW20V-DAWYOXHNVxzF:b-d2nyau) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2df90f2f327535487508c43052001f2265a1d5e9/sinks/dashboard/items/register.py#L2)):*

part of me would prefer the item list to be a member of dashboard, rather than its own thing. This means we could use encapsulation and have Register be a member class of dashboard. Although, this is a subjective idea
</blockquote></details>

IMO its cleaner to leave it as it is for separation of concerns, so dashboard doesn't need to care about how the dashitems are aggregated.

BluCodeGH (changes requested):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)* status: all files reviewed, 4 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 106 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKYcQvU14qUu3CY5e9j:-NKYcQvU14qUu3CY5e9k:bos0lrj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bb620c4d1e6af243f830a63e71c66fc3c8eb31b0/sinks/dashboard/dashboard.py#L106)):*
> ```Python
>         # First, make sure that the dash
>         # items within the docks are not
>         # Registryed with any series
> ```

Find and replace messed up this comment?
___
*[`sinks/dashboard/items/registry.py` line 1 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKYbmxU-vy-UB2A1YvZ:-NKYbmxU-vy-UB2A1Yv_:b6kpx2v) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bb620c4d1e6af243f830a63e71c66fc3c8eb31b0/sinks/dashboard/items/registry.py#L1)):*
> ```Python
> class Registry:
> ```

Imo keeping the class as `Register` is cleaner since the decorator (`@Register`) then makes more sense. Then the `get_items()` function could actually be outside of the class, so in the dashboard code you'd have `from items import registry` and `registry.get_items()`

Darth-Raazi (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 106 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKYcQvU14qUu3CY5e9j:-NK_NmWM9lpqSeTMT66X:bt2vr1b) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bb620c4d1e6af243f830a63e71c66fc3c8eb31b0/sinks/dashboard/dashboard.py#L106)):*

Find and replace messed up this comment?
</blockquote></details>

Fixed.
___
*[`sinks/dashboard/items/registry.py` line 1 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-NKYbmxU-vy-UB2A1YvZ:-NK_PB0V5HUMfUZ_tN04:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bb620c4d1e6af243f830a63e71c66fc3c8eb31b0/sinks/dashboard/items/registry.py#L1)):*

Imo keeping the class as `Register` is cleaner since the decorator (`@Register`) then makes more sense. Then the `get_items()` function could actually be outside of the class, so in the dashboard code you'd have `from items import registry` and `registry.get_items()`
</blockquote></details>

Done

BluCodeGH (approved):
:lgtm:

> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/86#-:-NK_RbYV7ZzjGCC2Ek1O:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @KavinSatheeskumar)

PR #89, opened 2022-12-31
Starting Omnibus

Darth-Raazi:
Tested on macOS and Windows. If this approach looks good, I'm thinking of adding a flag for every source (`--fake-ni`, `--fake-parsley`, `--parsley`, etc) and a profile to go with it. Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,5 +29,5 @@ dist/
 coverage.xml
 htmlcov
 
-*.sav
+*.json
 *.log
--- a/launcher.py
+++ b/launcher.py
@@ -22,7 +22,7 @@ profiles = {
 if test:
     profile = profiles["test"]
 else:
-    profile = profiles["false"]
+    profile = profiles["real"]
 
 p = []
 
--- a/launcher.py
+++ b/launcher.py
@@ -15,7 +15,7 @@ profiles = {
         ['python', 'sources/ni/main.py'],
     ],
     "Dashboard": [
-        ['python', 'sinks/globallog/main.py'],
+        #['python', 'sinks/globallog/main.py'],
         ['python', 'sinks/dashboard/main.py']
     ]
     # Add new profiles here
@@ -44,7 +44,7 @@ inp = ""
 
 # Run every file in the background
 for i in range(len(profile)):
-    p.append(Popen(profile[i]))
+    p.append(Popen(profile[i], stderr=PIPE))
 
     # Time delay cause CPU too fast and Omnibus 
     # needs time to setup communication channels
@@ -60,6 +60,9 @@ try:
         for process in p:
             if process.poll() != None:
                 raise Finished
-except (Finished, KeyboardInterrupt):
+except (Finished, KeyboardInterrupt, Exception):
     for i in p:
         i.terminate()
+        err = i.stderr.read().decode()
+        if err:
+            print(err)
--- a/launcher.py
+++ b/launcher.py
@@ -1,4 +1,6 @@
 from subprocess import Popen, PIPE
+from sys import platform
+from os import kill 
 import time, signal
 
 # Profiles for what parts of Omnibus should be run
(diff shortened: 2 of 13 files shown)


KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @BluCodeGH and @Darth-Raazi)

___
*[`launcher.py` line 6 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89#-NK_Xo0f197yVSmIdHLE:-NK_Xo0f197yVSmIdHLF:b2g0g12) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c540585a246393f8cd635e07140adbfbbc90cbc/launcher.py#L6)):*
> ```Python
> 
> parser = argparse.ArgumentParser()
> parser.add_argument('--test', action='store_true', help="run omnibus with FakeNI source")
> ```

Instead of using a bunch of command line arguments, I feel like a series of prompts might be easier (esp as we add more profiles!)

something like "select the option you want" and then it prints out all profiles and you enter a number from 1 to 10 or whatever
___
*[`launcher.py` line 15 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89#-NK_X5xW7F39_XzIC1Qa:-NK_X5xW7F39_XzIC1Qb:bndefm1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c540585a246393f8cd635e07140adbfbbc90cbc/launcher.py#L15)):*
> ```Python
>         ['python', 'sinks/dashboard/main.py']
>     ],
>     "real": [
> ```

practically speaking, this launcher is going to be most important for cold flows. This means that I'd prefer the following two options instead of "real"

"DAQ", which runs 
['python', '-m', 'omnibus']
['python', 'sources/ni/main.py']

and "Dashboard"
['python', 'sinks/dashboard/main.py']
['python', 'sinks/globallog/main.py']
___
*[`omnibus/server.py` line 81 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89#-NK_Y8ZYEbItdeyOAngn:-NK_Y8ZYEbItdeyOAngo:bn0bau0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c540585a246393f8cd635e07140adbfbbc90cbc/omnibus/server.py#L81)):*
> ```Python
>             monitor.recv_multipart()
>             count += 1
>         # if time.time() - t > 0.2:
> ```

why is this getting commented out?

BluCodeGH (changes requested):
I like the approach of putting the processes in an array we can do things with, but there is some desired functionality that is currently not implemented:
* We should kill everything if any one of the processes exits, not just the last one, or on a KeyboardInterrupt.
* Input to the console should be forwarded to _all_ of the launched processes (so we can enter the server IP if required)
* We should find a way to let multiple processes show their output on the console, for example the server msgs/sec and the globallog "Logging...". I find seeing the msgs/sec very useful for debugging connection issues and we want to make sure stuff like logging doesn't crash unexpectedly.
* If one of the processes exits, its output (and stderr) should be printed to the console (we want to be able to see the traceback / debug crashes)
* We should kill with sigkill, wait a second, and then sigterm any remaining processes so they have time to close files / flush buffers / etc.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89#-:-NKd2Mf1-0exkDyl5yNy:b3htdu5)* status: all files reviewed, 3 unresolved discussions (waiting on @Darth-Raazi)

BluCodeGH (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/89)* status: all files reviewed, 3 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

PR #91, opened 2023-01-11
plot multiple series in one plot

yxyyeah:
屏幕截图 2023-01-10 190814
Trying to achieve this

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -61,7 +61,16 @@ class Dashboard(QtWidgets.QWidget):
         def prompt_and_add(i):
             def ret_func():
                 props = registry.get_items()[i].prompt_for_properties(self)
-                if props:
+                #value returned from PlotDashItem
+                if isinstance(props, list):
+                    #plot separately
+                    if props[0][1]:
+                        self.add(registry.get_items()[i]([[[i], props[1]] for i in props[0][0]]))
+                    #plot multi-series in one plot
+                    else:
+                        self.add(registry.get_items()[i](props))
+                #value returned from CanMsgTableDashItem
+                elif props:
                     self.add(registry.get_items()[i](props))
             return ret_func
 
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -61,15 +61,16 @@ class Dashboard(QtWidgets.QWidget):
         def prompt_and_add(i):
             def ret_func():
                 props = registry.get_items()[i].prompt_for_properties(self)
-                #value returned from PlotDashItem
+                # value returned from PlotDashItem
                 if isinstance(props, list):
-                    #plot separately
+                    # plot separately
                     if props[0][1]:
-                        self.add(registry.get_items()[i]([[[i], props[1]] for i in props[0][0]]))
-                    #plot multi-series in one plot
+                        for j in range(len(props[0][0])):
+                            self.add(registry.get_items()[i]([[props[0][0][j]], props[1][j]]))
+                    # plot multi-series in one plot
                     else:
-                        self.add(registry.get_items()[i](props))
-                #value returned from CanMsgTableDashItem
+                        self.add(registry.get_items()[i]([props[0][0], props[1]]))
+                # value returned from CanMsgTableDashItem
                 elif props:
                     self.add(registry.get_items()[i](props))
             return ret_func
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -61,11 +61,7 @@ class Dashboard(QtWidgets.QWidget):
         def prompt_and_add(i):
             def ret_func():
                 props = registry.get_items()[i].prompt_for_properties(self)
-                # value returned from PlotDashItem
-                if isinstance(props, list):
-                    self.add(registry.get_items()[i]([props[0], props[1]]))
-                # value returned from CanMsgTableDashItem
-                elif props:
+                if props:
                     self.add(registry.get_items()[i](props))
             return ret_func
 
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -63,13 +63,7 @@ class Dashboard(QtWidgets.QWidget):
                 props = registry.get_items()[i].prompt_for_properties(self)
                 # value returned from PlotDashItem
                 if isinstance(props, list):
-                    # plot separately
-                    if props[0][1]:
-                        for j in range(len(props[0][0])):
-                            self.add(registry.get_items()[i]([[props[0][0][j]], props[1][j]]))
-                    # plot multi-series in one plot
-                    else:
-                        self.add(registry.get_items()[i]([props[0][0], props[1]]))
+                    self.add(registry.get_items()[i]([props[0], props[1]]))
                 # value returned from CanMsgTableDashItem
                 elif props:
                     self.add(registry.get_items()[i](props))
--- a/sinks/dashboard/items/can_message_table.py
+++ b/sinks/dashboard/items/can_message_table.py
@@ -187,7 +187,7 @@ class CanMsgTableDashItem(DashboardItem):
     def get_props(self):
         return self.props
 
-    def on_data_update(self, canSeries):
+    def on_data_update(self, stream, canSeries):
         message = canSeries[1]
         if message["board_id"] in self.message_dict:
             self.message_dict[message["board_id"]].update_with_message(message)
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -23,7 +23,7 @@ class PlotDashItem(DashboardItem):
         super().__init__()
 
         self.size = config.GRAPH_RESOLUTION * config.GRAPH_DURATION
-        self.last = 0
+        self.last = {}
 
         # storing the series name as key, its time and points as value
         # since each PlotDashItem can contain more than one curve
@@ -43,8 +43,8 @@ class PlotDashItem(DashboardItem):
         self.series = self.props[0]
 
         # subscribe to stream dictated by properties
-        for i in range(len(self.series)):
-            publisher.subscribe(self.series[i], self.on_data_update)
+        for series in self.series:
+            publisher.subscribe(series, self.on_data_update)
 
         # a default color list for plotting multiple curves
         # yellow green cyan white blue magenta
@@ -60,11 +60,12 @@ class PlotDashItem(DashboardItem):
         # storing the series name as key, its plot object as value
         # update all curves every time on_data_update() is called
         self.curves = {}
-        for i in range(len(self.series)):
-            curve = self.plot.plot([], [], pen=self.color[i], name=self.series[i])
-            self.curves[self.series[i]] = curve
-            self.times[self.series[i]] = np.zeros(self.size)
-            self.points[self.series[i]] = np.zeros(self.size)
+        for i,series in enumerate(self.series):
+            curve = self.plot.plot([], [], pen=self.color[i], name=series)
+            self.curves[series] = curve
+            self.times[series] = np.zeros(self.size)
+            self.points[series] = np.zeros(self.size)
+            self.last[series] = 0
 
         if self.limit is not None:
             self.warning_line = self.plot.plot([], [], brush=(255, 0, 0, 50), pen='r')
@@ -80,7 +81,7 @@ class PlotDashItem(DashboardItem):
         channel_and_series = prompt_user(
             self,
             "Data Series",
-            "Select up to 6 serie(s) you wish to plot",
+            "Select up to 6 series you wish to plot",
             "checkbox",
             publisher.get_all_streams(),
         )
@@ -106,19 +107,15 @@ class PlotDashItem(DashboardItem):
         desc = payload[2] if (len(payload) > 2) else ""
 
         # time should be passed as seconds, GRAPH_RESOLUTION is points per second
-        if time - self.last < 1 / config.GRAPH_RESOLUTION:
+        if time - self.last[stream] < 1 / config.GRAPH_RESOLUTION:
             return
 
-        if self.last == 0:  # is this the first point we're plotting?
+        if self.last[stream] == 0:  # is this the first point we're plotting?
             # prevent a rogue datapoint at (0, 0)
-            # here we are filling in all curves with the same initial value, which is not accurate
-            # but once the curve gets its own update, the following values will be accurate
-            for v in self.times.values():
-                v.fill(time)
-            for v in self.points.values():
-                v.fill(point)
+            self.times[stream].fill(time)
+            self.points[stream].fill(point)
 
-        self.last += 1 / config.GRAPH_RESOLUTION
+        self.last[stream] += 1 / config.GRAPH_RESOLUTION
 
         # add the new datapoint to the end of the corresponding stream array, shuffle everything else back
         self.times[stream][:-1] = self.times[stream][1:]
@@ -140,9 +137,8 @@ class PlotDashItem(DashboardItem):
             # set the red tint
             self.warning_line.setFillLevel(max_point*2)
 
-        # plot the data curves
-        for k, v in self.curves.items():
-            v.setData(self.times[k], self.points[k])
+        # update the data curve
+        self.curves[stream].setData(self.times[stream], self.points[stream])
 
         # value readout in the title
         # avg values
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -40,16 +40,16 @@ class PlotDashItem(DashboardItem):
 
         # save props as a field
         self.props = props
-        self.series = self.props[0][0]
+        self.series = self.props[0]
 
         self.curve_count = 0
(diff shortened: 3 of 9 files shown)


KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @yxyyeah)

___
*[`sinks/dashboard/dashboard.py` line 63 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdEsfq1W_kxPYXaLcM:-NLdEsfq1W_kxPYXaLcN:buqpahk) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sinks/dashboard/dashboard.py#L63)):*
> ```Python
>         def prompt_and_add(i):
>             def ret_func():
>                 props = registry.get_items()[i].prompt_for_properties(self)
> ```

A couple of thoughts
1. I think that adding the new feature "plot separately" forces this code here to become to specific to particular dash items (IE, the plot). So it may be best to leave it out for now and come back to this in a separate PR
2. the properties for plot_dash_item is turning into a rather large list where each field means a different thing. I think the code would be easier to read if we turned it into a dict. 
    - However, since we are removing the "plot separately" button for now, we the props should be a fair bit simpler, meaning it can stay a list until we add that back
___
*[`sinks/dashboard/utils.py` line 57 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdGD4b-sAd6ONhGh2v:-NLdGD4b-sAd6ONhGh2w:b-ibl8rs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sinks/dashboard/utils.py#L57)):*
> ```Python
>         # set up a dialog template
>         dialog_template = QtWidgets.QInputDialog()
>         if prompt_type == "text" or prompt_type == "items":
> ```

This if-statement here is case-splitting on the same value prompt_type, so we don't need to put it into the else case of this if-statement. Instead, repeat the line dialog_template = QtWidgets.QInputDialog() inside each of the cases in the if-statement below
___
*[`sinks/dashboard/items/plot_dash_item.py` line 53 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdKuiC7eqazW9nHIJt:-NLdKuiC7eqazW9nHIJu:b-djw1wq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sinks/dashboard/items/plot_dash_item.py#L53)):*
> ```Python
> 
>         # a default color list for plotting multiple curves
>         self.color = ['w', 'g', 'b', 'm', 'y', 'c']
> ```

It'd be awesome if we could make this the colour list ['y', 'r', 'g', 'b', 'w', 'c']

I am a tad sentimental about the yellow plots, so I wanna keep them lol. Its the team colour!
___
*[`sinks/dashboard/items/plot_dash_item.py` line 116 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdJmlI5GK9m98A0aj9:-NLdJmlI5GK9m98A0ajA:ba341v8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sinks/dashboard/items/plot_dash_item.py#L116)):*
> ```Python
>                 cancelText="No Threshold"
>             )
>             props = [channel_and_series, threshold_input]
> ```

We might want to add a check for the length of "channel_and_series" to ensure that it isn't too long (else the dashboard just ends up erroring out the ass.) If the user selects more series than we have colours, then just only take the first few selected series. We can add a not about this in the prompt.
___
*[`sources/ni/config.py` line 7 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdLC967XJlsTstBsP2:-NLdLC967XJlsTstBsP3:bg9d06z) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sources/ni/config.py#L7)):*
> ```Python
> 
> 
> def setup():
> ```

Was this just changed by the format.sh tool? If so, then no worries

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @yxyyeah)

___
*[`sinks/dashboard/items/plot_dash_item.py` line 30 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLdLlyeAhXF4E0YzAyb:-NLdLlyeAhXF4E0YzAyc:byea2ou) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/4b94822b7369cd64359e0ceb42390963cb699a80/sinks/dashboard/items/plot_dash_item.py#L30)):*
> ```Python
>         # "size" of running average
>         self.avgSize = config.RUNNING_AVG_DURATION * config.GRAPH_RESOLUTION
>         self.time_offset = 0
> ```

As per your question, you can feel free to remove any unused variables

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @yxyyeah)

___
*[`sinks/dashboard/dashboard.py` line 64 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLmujPf7vkEtz1O7lsj:-NLmujPf7vkEtz1O7lsk:blrotce) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fd69a6095c5efd0fe202f7f391c76be9786fd308/sinks/dashboard/dashboard.py#L64)):*
> ```Python
>             def ret_func():
>                 props = registry.get_items()[i].prompt_for_properties(self)
>                 # value returned from PlotDashItem
> ```

This piece of code can be returned back to how it was without issue

BluCodeGH (changes requested):
Looks like a promising start, although I do really think we need to work on a better generic dashboarditem properties system (not in this PR) since all these dialogues are getting a bit extreme.

> Reviewed 3 of 7 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-:-NM-tRE_3IGnuy4rTUwl:b5ionq0)* status: all files reviewed, 9 unresolved discussions (waiting on @KavinSatheeskumar and @yxyyeah)

___
*[`sinks/dashboard/utils.py` line 42 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-e_016K7azAXDmKTd:-NM-e_016K7azAXDmKTe:bm92rhr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/utils.py#L42)):*
> ```Python
>         if dia.exec():
>             selected_items = []
>             for i in range(items_length):
> ```

Probably cleaner to iterate over `enumerate(items)`, then you don't need the `list()` call below.
___
*[`sinks/dashboard/utils.py` line 46 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-dotD589LexAXxgY_:-NM-dotD589LexAXxgYa:b-mvlthu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/utils.py#L46)):*
> ```Python
>                     selected_items.append(list(items)[i])
>             return selected_items
>     else:
> ```

Rather than an `else` and indenting the rest of the code, we can explicitly `return None` if the `dia.exec()` returns false. Then the checkbox prompt type always returns something so you don't need an else case.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 29 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-jv00BzvK0DlmV_FE:-NM-jv00BzvK0DlmV_FF:b-h4n031) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/1e899ae69b5bb3d58b4ade37a8e8be534d244cb8/sinks/dashboard/items/plot_dash_item.py#L29)):*
> ```Python
>         self.times = np.zeros(self.size)
>         self.points = np.zeros(self.size)
>         self.sum = 0  # sum of stream
> ```

Having a running sum to calculate the average from is an important performance feature. You should have a `self.sums = {}` and adapt the code as necessary.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 46 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-kPYx7zCeW-evD1Tw:-NM-kPYy81htI-g8zlhb:b-9i42jp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L46)):*
> ```Python
> 
>         # subscribe to stream dictated by properties
>         for i in range(len(self.series)):
> ```

`for series in self.series:`
___
*[`sinks/dashboard/items/plot_dash_item.py` line 63 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-karu7rAXxxTbEDmg:-NM-karu7rAXxxTbEDmh:bbhv5pz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L63)):*
> ```Python
>         # update all curves every time on_data_update() is called
>         self.curves = {}
>         for i in range(len(self.series)):
> ```

`for i, series in enumerate(self.series):`
___
*[`sinks/dashboard/items/plot_dash_item.py` line 83 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-qjHkD-oxVVCu38EQ:-NM-qjHkD-oxVVCu38ER:b-919ihp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L83)):*
> ```Python
>             self,
>             "Data Series",
>             "Select up to 6 serie(s) you wish to plot",
> ```

"Select up to 6 series you wish to plot", the singular of series is series
___
*[`sinks/dashboard/items/plot_dash_item.py` line 119 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-s2yr-MkFa7BbpA4p:-NM-s2yr-MkFa7BbpA4q:b3ksheq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L119)):*
> ```Python
>                 v.fill(time)
>             for v in self.points.values():
>                 v.fill(point)
> ```

This will fill every series with the first point of the first series measured. You need a per-series `self.last` and to only fill with the initial value one series at a time.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 145 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-tAi830ac22pX7WLU:-NM-tAi830ac22pX7WLV:b-1f1b78) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L145)):*
> ```Python
>         # plot the data curves
>         for k, v in self.curves.items():
>             v.setData(self.times[k], self.points[k])
> ```

You only need to set the data on the curve that we just got a new point for

yxyyeah (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91)* status: 5 of 7 files reviewed, 9 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`sinks/dashboard/utils.py` line 42 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-e_016K7azAXDmKTd:-NMJkLpA9g5800zaVfww:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/utils.py#L42)):*

Probably cleaner to iterate over `enumerate(items)`, then you don't need the `list()` call below.
</blockquote></details>

Done.
___
*[`sinks/dashboard/utils.py` line 46 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-dotD589LexAXxgY_:-NMJk9Sb4xJUPdEM1EB7:barrzhr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/utils.py#L46)):*

Rather than an `else` and indenting the rest of the code, we can explicitly `return None` if the `dia.exec()` returns false. Then the checkbox prompt type always returns something so you don't need an else case.
</blockquote></details>

For this one I think the if statement is determine the prompt type. The if one is checkbox and the else block is the other 2 types. I think categorizing the types in this way is more clear?
___
*[`sinks/dashboard/items/plot_dash_item.py` line 46 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-kPYx7zCeW-evD1Tw:-NMJkL-M9JF10m6_ORYy:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L46)):*

`for series in self.series:`
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 63 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-karu7rAXxxTbEDmg:-NMJkKFb816nX_eNnTQv:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L63)):*

`for i, series in enumerate(self.series):`
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 83 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-qjHkD-oxVVCu38EQ:-NMJkJaVD8ZdtFwd5OLI:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L83)):*

"Select up to 6 series you wish to plot", the singular of series is series
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 119 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-s2yr-MkFa7BbpA4p:-NMJkHb_5BP8CdStMvus:b-8y1tq0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L119)):*

This will fill every series with the first point of the first series measured. You need a per-series `self.last` and to only fill with the initial value one series at a time.
</blockquote></details>

This field is never used in the function, and I also deleted some other unused variables. I think we can get to this later when we need them?
___
*[`sinks/dashboard/items/plot_dash_item.py` line 145 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NM-tAi830ac22pX7WLU:-NMJkNGb45Y8_eUCPXVm:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/5c4c83378decc85ec06761980094dbf658fcde0e/sinks/dashboard/items/plot_dash_item.py#L145)):*

You only need to set the data on the curve that we just got a new point for
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 64 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-NLmujPf7vkEtz1O7lsj:-NMJkMruCSX3wmY3wtLm:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/fd69a6095c5efd0fe202f7f391c76be9786fd308/sinks/dashboard/dashboard.py#L64)):*

This piece of code can be returned back to how it was without issue
</blockquote></details>

Done.

BluCodeGH (approved):
:lgtm:

> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-:-NNDG3WjDHRVCy9aQIP0:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @KavinSatheeskumar)

KavinSatheeskumar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/91#-:-NNEePP1-H32LT2eEZrW:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @yxyyeah)

PR #95, opened 2023-02-24
Allow omnibus to send CAN messages

itslinotlie:
We're back baby
[clickup link](https://app.clickup.com/t/37c2nya)
I made [this diagram](https://drive.google.com/file/d/1zixXpazHdTN_Lg0c4ZYoL6dzV93u855q/view?usp=sharing), it might be helpful to others
(ignore the commits, I messed up while rebasing a couple of times...)

### Cool UI features
- when you type something into a text field and said text field reaches full capacity, the focus is set to the next widget in the CAN message
  - if this next CAN message is a text field, that means you can continue typing without having to click anything
  - if this next CAN message is a drop down, that means you can use the keyboard arrow keys to go through the dropdown items without having to click anything
- when you press backspace on a text field and there are no more characters in said text field, the focus is set to the previous widget in the CAN message (and the last character from this previous widget is deleted, if possible)
  - if this previous CAN message is a text field, that means you can continue typing without having to click anything
  - if this previous CAN message is a drop down, that means you can use the keyboard arrow keys to go through the dropdown
 - you can press tab / shift+tab to change the widget focus (ie. if you're currently changing the msg_type, you can press tab to move the focus of the widget to board_id, where you can then change that value)
 - you can press enter/return whenever the CAN sender dashboard item is in focus to send the CAN message
   - TLDR; you don't really have to use the mouse to navigate around the dashboard item 
- we have visual feedback when the SEND button is pressed!
  - for fields that contain invalid data, the bad fields are pulsed consecutively in quick succession and a label with the error appears below the widget that failed
  - if all fields are valid, all the fields are pulsed once 

### demo video
https://user-images.githubusercontent.com/56371758/233762963-e711afeb-b131-47b3-9ff6-c284de5cf450.mp4

### Issues that (might) need to be addressed in the future

- cannot scroll over a dropdown field, which is a native Windows behaviour. The problem is that with the current way our dashboard is set up, the QGraphicsView consumes all the scrolling wheel events. It's possible to see what widget the mouse is currently hovering over, but that will only tell you you're hovering over the dashboard item. The following was my poc code that doesn't really work because I wasn't able to figure out how to type match the dashboard widget with a specific class (without creating an instance of said class)
  - ```
        relative_pos = QCursor.pos() # returns cursor position relative to the scene
        absolute_pos = self.mapFromGlobal(relative_pos) # returns the absolute cursor position
        hovered_widget = self.itemAt(absolute_pos) # uses QGraphicView's itemAt to determine the widget being hovered

        if hovered_widget is not None:
            nested_widget = hovered_widget.widget() # returns dashboard widget
            # find out some way to relate the dashboard widget to if the widget is scrollable (in which case, pass this event down)
            return super().wheelEvent(event)
     ```

- zooming in/out of the dashboard messes with the QComboBox's dropdown fields
![image](https://user-images.githubusercontent.com/56371758/233528451-064f6fca-9a0a-4012-9c2f-c76f167a4195.png)
  - If we want a robust solution, there probably needs to be a signal that's emitted from the dashboard whenever the view scaling changes so that all the dashboard widgets can update themselves accordingly

- dropdown isn't customizable on Mac (can't do anything about it, see [here](https://doc.qt.io/qt-5/qcombobox.html#maxVisibleItems-prop))
![image](https://user-images.githubusercontent.com/56371758/233528706-aae824f6-c7a5-4022-8f77-4ab65ee14945.png)
  - this means that Mac users will see a huge list of fields in their dropdown
  - for some reason, you also can't scroll through the widgets on Mac. On Windows, if you have the mouse hovered over the dropdown (the dropdown list isn't expanded) and scroll, the wheel event is passed and the dropdown fields change ([default behaviour](https://doc.qt.io/qt-5/qcombobox.html#maxVisibleItems-prop) doesn't apply to Mac I guess ¯\_(ツ)_/¯)

- I don't understand why there's so much padding/margin between widgets
![image](https://user-images.githubusercontent.com/56371758/233529501-2df67928-0a5c-40d8-8d60-1f83353c515d.png)
  - the widgets are next to each other in the layout manager, I've tried ```self.layout_manager.setContentsMargins(0, 0, 0, 0)``` and ```self.layout_manager.setSpacing(0)``` and ```self.layout_manager.setColumnStretch(self.widget_index, 0)``` but the width between widgets isn't addressed. This issue of large padding applies to only CAN messages with a small number of fields

- it would be really nice if someone could create keyboard shortcuts for adding dashboard items because then, you can interact with Omnibus without having to actually click on anything

- I tried really hard for this one, and I really wanted it to work, but I cannot figure it out. Most widgets come with some slots that you can connect to (observer pattern), but oftentimes, we need to customize the signals that are emitted (eg. finding out when a backspace is pressed from a text field). Currently, this is done by creating a class that inherits from the widget that needs to be customized, and I wanted to centralize these signal/event logic into a centralized class. However, I think improving/customizing Omnibus's observer pattern is an endeavour on its own. The problem is that a QGraphicsView (see the diagram at the top) translates QWheelEvents into QGraphicsSceneWheelEvents and we lose out on information on where the mouse event originated (we handle things differently if it comes from a mouse or touchpad). This means that we have to inherit from QGraphicsView to intercept these wheel events before they get translated, which kind of defeats the whole purpose of a centralized event class...

- unable to show QComboBox focus border highlight when it gets focus from the keyboard (if you were to click on it, you would see a blue border indicating that the widget has focus, but I couldn't figure out a way to do it for the keyboard). You can technically change the CSS to ```dropdown.setStyleSheet("QComboBox:focus { border: 1px solid #0079d6; }")``` but then you override the other QComboBox:Focus CSS, which is probably worse.

- okay this point is me being lazy (but to be fair, there was a lot of new things thrown at me since the start of this PR), but this is somewhat important. We need to set up a slot/signal connection between the dashboard and the dash items so that dash items can communicate to the dashboard when their layout has changed so that the dashboard can then update the QGraphicsRectItem of that dash item.
![image](https://user-images.githubusercontent.com/56371758/233762162-e5af21d8-8d78-437e-8f28-eed78de7ad9f.png)
  - The part outside the dotted lines is un-selectable since the QGraphicsRectItem is out of date.


<!-- Reviewable:start -->
- - -
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95)
<!-- Reviewable:end -->


sinks/dashboard/items/can_sender/can_sender.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, props=None):
+        super().__init__()
+        self.props = props
+
+        self.layout_manager = QtWidgets.QGridLayout(self)
+        self.setLayout(self.layout_manager)
+
+        self.canlib_info = CanlibMetadata("can_sender_data.txt")
+        self.sender_thing = Sender()
```
itslinotlie, line 38: I am god awful at naming variables and the name ```self.sender``` conflicts with Sender(), this probably needs a better name

sinks/dashboard/items/can_sender/can_sender.py
```diff
@@ -0,0 +1,217 @@
...
+    __number_of_bytes = 8
+
+    def __init__(self, props=None):
+        super().__init__()
+        self.props = props
+
+        self.layout_manager = QtWidgets.QGridLayout(self)
+        self.setLayout(self.layout_manager)
+
+        self.canlib_info = CanlibMetadata("can_sender_data.txt")
```
itslinotlie, line 37: This will be gone after the parsley rewrite

sinks/dashboard/items/can_sender/can_sender.py
```diff
@@ -0,0 +1,217 @@
...
+    def parse_data(self):
+        msg_type = self.message_type.currentText()
+        msg_data = b""
+        amt_of_data = len(self.canlib_info.get_msg_data(msg_type))
+        for i in range(amt_of_data):
+            # padding text with 0 until len = 2
+            msg_data += bytes.fromhex(self.line_edits[i].text().zfill(2))
+
+        msg_type = mt.msg_type_hex[msg_type]
+        msg_board = 0
```
itslinotlie, line 119: QUESTION: do we have a board_id for omnibus?

itslinotlie:
I didn't realize the | in DEBUG_MSG was referring to a half byte, that will be addressed after the parsley rewrite

KavinSatheeskumar (dismissed):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95)* status: 0 of 5 files reviewed, 10 unresolved discussions (waiting on @itslinotlie)

___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 119 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NP0e2DXdeyzM8jE7jeU-r1-119:-NPG2dxv2KNmvdPSZfQV:b-66hdmp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/85a1f37e8730052ca6fa511e8f3beffa30a0f134/sinks/dashboard/items/can_sender/can_sender.py#L119)):*

QUESTION: do we have a board_id for omnibus?
</blockquote></details>

There exists a board ID "DUMMY" which may be useful for this purpose
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 12 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3zfH-ue_o23Q3oZS:-NPG3zfH-ue_o23Q3oZT:bsnk203) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L12)):*
> ```Python
> 
> @Register
> class CanSender(DashboardItem):
> ```

What might be nice is having the timestamp fields auto-filled. (So we don't need to worry about stuffing them with garbage data) If that isn't too much work
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 126 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG4MXc37WFQ_9ULX0C:-NPG4MXc37WFQ_9ULX0D:b-ltcobs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L126)):*
> ```Python
>          being sent and requires that there be a [data][time] in the message object so...
>         """
>         formatted_data = {'data': {'time': -1}, 'message': (msg_sid, msg_data)}
> ```

We could just use time.time() for the time stamp. We could also use that for the timestamp on the can messages we send
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 196 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3X2q4ir1LOmf9Tep:-NPG3X2q4ir1LOmf9Teq:b3khr4d) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L196)):*
> ```Python
> 
>     def get_props(self):
>         return self.props
> ```

We don't need the field self.props, just make this also return None (or maybe True, test out the save and load to make sure whatever value you use works)
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 199 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3OrX8yh7nkmewtbc:-NPG3OrX8yh7nkmewtbd:b-or5ca0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L199)):*
> ```Python
> 
>     def on_delete(self):
>         publisher.unsubscribe_from_all(self.on_data_update)
> ```

We can just make this pass
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 202 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3QqWBZ4xJCO0gZMC:-NPG3QqWBZ4xJCO0gZMD:b88o8v5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L202)):*
> ```Python
> 
>     def prompt_for_properties(self):
>         return True
> ```

We can just make this return None
___
*[`sinks/dashboard/items/can_sender/canlib_metadata.py` line 1 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG1Bw16BWWSu_8W7YX:-NPG1Bw16BWWSu_8W7YY:b-jzamo8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/canlib_metadata.py#L1)):*
> ```Python
> import re
> ```

When we re-write parsely, we should make this "info" dictionary into a hardcoded constant rather than having it in a textfile and then parsing the text file.
___
*[`sinks/dashboard/items/can_sender/canlib_metadata.py` line 4 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG1rmt86D6QigsP0tY:-NPG1rmt86D6QigsP0tZ:bz6qgsn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/canlib_metadata.py#L4)):*
> ```Python
> import os
> 
> class CanlibMetadata:
> ```

I know this is getting re-written. However, I think that this dosen't need to be a class in and of itself. It can just be a function that returns a dictionary. (which will eventually be hard coded, see above)

BluCodeGH (changes requested):
Looking good, I've gone through and left a bunch of nitpicky comments. I don't think we should merge this until we've finished splitting out parsley since the current comment parsing stuff is pretty messy and I don't want to give it a chance to persist :stuck_out_tongue: .

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-:-NPTY09t3ol7mrfR7qZY:b7ixgmg)* status: all files reviewed, 26 unresolved discussions (waiting on @itslinotlie and @KavinSatheeskumar)

___
*[`poc.py` line 99 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQC1m4NrDVaiE9zhZ:-NPTQC1m4NrDVaiE9zh_:bf0pnhs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L99)):*
> ```Python
>         return self.map_val_num[value].to_bytes((self.length + 7) // 8, 'big')
> 
> class Ascii(Field):
> ```

ASCII is an acronym (all uppercase)
___
*[`poc.py` line 108 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTPvEC313f89dKvK0k:-NPTPvEC313f89dKvK0l:b2b12b9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L108)):*
> ```Python
>     def encode(self, value):
>         # i need to go test to ensure my sanity
>         if self.length < len(value):
> ```

This is comparing self.length (in bits) with the string length (in characters). Probably easiest to just try to encode it and check the length after that.
___
*[`poc.py` line 111 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQ3S19wZyqrBxHuq2:-NPTQ3S22_SocW6XwlVS:b-nxciyj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L111)):*
> ```Python
>             raise ValueError(f"String {value} is too large for {self.length}")
>         if not value.isascii():
>             raise UnicodeEncodeError(f"Value contains non-ascii characters")
> ```

We should only raise ValueErrors so they are easy to catch and handle.
___
*[`poc.py` line 114 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTPPCA7QPgTAFXhSZU:-NPTPPCA7QPgTAFXhSZV:b-pqq8lz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L114)):*
> ```Python
>         return value.encode('ascii')
> 
> TIMESTAMP_2 = Numeric("time", 16, scale=0.001, signed=True)
> ```

TIMESTAMP_2 is just a two-byte version of the usual timestamp to save space (so it rolls over faster), it shouldn't be signed.
___
*[`poc.py` line 118 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTPZOaAOp-0JVqQjpL:-NPTPZOaAOp-0JVqQjpM:bt3u9ej) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L118)):*
> ```Python
> 
> FIELDS = {
>     "GENERAL_CMD": [TIMESTAMP_3, Enum("command", 8, mt.gen_cmd_hex)],
> ```

Just a note that we should remove the _hex suffix from all the dictionaries in message_types, and remove all the reversed dictionaries, since Enum calculates the reverse automatically.
___
*[`poc.py` line 158 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQV0S4x3s0OdyMbna:-NPTQV0S4x3s0OdyMbnb:bttzifv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/poc.py#L158)):*
> ```Python
> 
>     result = {"msg_type": msg_type, "board_id": board_id, "data": {}}
>     for field in FIELDS[msg_type]:
> ```

We should add a check for unrecognized message types that just returns `"data": {"unknown": bytes_of_message}`
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 38 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NP0e2DXdeyzM8jE7jeU-r1-38:-NPTS9Lx-B-9gNSpsPEW:b-u0wlfi) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/85a1f37e8730052ca6fa511e8f3beffa30a0f134/sinks/dashboard/items/can_sender/can_sender.py#L38)):*

I am god awful at naming variables and the name ```self.sender``` conflicts with Sender(), this probably needs a better name
</blockquote></details>

`self.omnibus_sender`.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 119 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NP0e2DXdeyzM8jE7jeU-r1-119:-NPTURiLB3Fr0BwgajTj:b65plvv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/85a1f37e8730052ca6fa511e8f3beffa30a0f134/sinks/dashboard/items/can_sender/can_sender.py#L119)):*

There exists a board ID "DUMMY" which may be useful for this purpose
</blockquote></details>

Yeah 0 (DUMMY) works for now, we can probably add a dedicated one in the future.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 12 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3zfH-ue_o23Q3oZS:-NPI6hsm8DFkaM5PhUKt:b-s6ug1u) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L12)):*

What might be nice is having the timestamp fields auto-filled. (So we don't need to worry about stuffing them with garbage data) If that isn't too much work
</blockquote></details>

The timestamp fields are only 2 or 3 bytes so the current unix time isn't applicable. They can almost always be safely populated as zeros, which we should do automatically.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 126 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG4MXc37WFQ_9ULX0C:-NPTVKrNBMpivqNEs7_P:b-5s3tiw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L126)):*

We could just use time.time() for the time stamp. We could also use that for the timestamp on the can messages we send
</blockquote></details>

The error here is because the CAN parser (very end of `dashboard/parsers.py`) listens to the "CAN" channel. The fix here is to just change it to only listen to "CAN/Parsley" channel since it evidently depends on Parsley's formatting.

Also just noting that we should let the user control the timestamp bytes so they should be zero-initialized, not with `time.time()`.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 196 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3X2q4ir1LOmf9Tep:-NPTRnUM-8U3hZOSSS8b:b9vvnfq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L196)):*

We don't need the field self.props, just make this also return None (or maybe True, test out the save and load to make sure whatever value you use works)
</blockquote></details>

Yeah just `return True` here should do fine.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 199 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3OrX8yh7nkmewtbc:-NPTRaVN-M6O1sxVwU4U:b-7yy70l) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L199)):*

We can just make this pass
</blockquote></details>

No, just passing would be a memory leak since the callback would still be stored by the publisher. The current implementation is correct.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 202 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3QqWBZ4xJCO0gZMC:-NPTRMYZ91VjVah_T_D8:bqv9siz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L202)):*

We can just make this return None
</blockquote></details>

No, if prompt_for_properties returns a false-y value then the dashboard cancels widget creation. The current implementation is correct.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 26 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQtDC1kZ3WT2gbvhR:-NPTQtDC1kZ3WT2gbvhS:b-ha0jt4) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L26)):*
> ```Python
>     lbl      = describes the type of data in that column
>     """
>     __pulse_occurances = 3
> ```

NUM_PULSES
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 28 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQqP34KvyzxdG0bbo:-NPTQqP34KvyzxdG0bbp:b4jug0f) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L28)):*
> ```Python
>     __pulse_occurances = 3
>     __pulse_frequency = 100 #in ms
>     __number_of_bytes = 8
> ```

Constants should be ALL_CAPS_CASE. Probably "MAX_MESSAGE_BYTES" is a better name than NUMBER_OF_BYTES.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 32 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTRsRS8V-GOsybpzNw:-NPTRsRS8V-GOsybpzNx:b4leij7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L32)):*
> ```Python
>     def __init__(self, props=None):
>         super().__init__()
>         self.props = props
> ```

Props field is unnecessary like Kavin mentioned below.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 43 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTStcu5gvU7qk3hlzt:-NPTStcu5gvU7qk3hlzu:b1n9fkm) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L43)):*
> ```Python
>         self.setupWidgets()
>         self.logicfyWidgets()
>         self.placeWidgets()
> ```

I would rename `setup` -> `create` and `logicfy` -> `setup`. Also `snake_case` for function names please.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 106 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTTmIF5KAYHFeEOvOB:-NPTTmIF5KAYHFeEOvOC:b-bpxx09) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L106)):*
> ```Python
>     def send_can_message(self):
>         # there were invalid field(s) and we pulsed, dont send message
>         if self.pulse_widgets():
> ```

`pulse_widgets` doesn't make sense as a name in this context. Probably best to rename to something like `check_message_incomplete`.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 146 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTX21x88w1rpV6DCd9:-NPTX21x88w1rpV6DCdA:b-22wihz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L146)):*
> ```Python
>             self.timer.stop()
> 
>     def get_invalid_bytes(self):
> ```

`get_invalid_inputs` would be better since it includes the message type field.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 153 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTWIUY47lkz8mUsvxu:-NPTWIUY47lkz8mUsvxv:b-o54rnn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L153)):*
> ```Python
>             bad_indexes.append(-1)
>         for i in range(len(msg_data_info)):
>             if not self.line_edits[i].text():
> ```

If I try to type in something invalid like "hi" into one of the data byte inputs, will this catch that? I see that there is the QT regex validator above but will that let me type in something like that and just be mad about it or will it actually prevent me from typing invalid stuff in?
___
*[`sources/parsley/main.py` line 6 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTXTRTERsqkTsUmVIi:-NPTXTRU0Q4tZsmQt-wK:bmt5tu2) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sources/parsley/main.py#L6)):*
> ```Python
> from omnibus import Sender, Receiver
> import parsley
> import sources.parsley.message_types as mt
> ```

You should be able to just `import message_types as mt`.
___
*[`sources/parsley/main.py` line 16 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTXeOP5DKl3eKpyNUx:-NPTXeOP5DKl3eKpyNUy:b1tp9y8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sources/parsley/main.py#L16)):*
> ```Python
> 
>     def read(self):
>         if self.port == "-":
> ```

A better way to do this check would be to set `self.serial = None` in `__init__` and then check that here, so that if we change the string that represents using the console we only need to change it once.
___
*[`sources/parsley/main.py` line 21 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTX_3l09IhAQMvpXAH:-NPTX_3l09IhAQMvpXAI:b-k0he1p) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sources/parsley/main.py#L21)):*
> ```Python
> 
>     def write(self, msg):
>         self.serial.write(msg)
> ```

This is going to crash if we passed in "-", try printing to stdout?

itslinotlie (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95)* status: all files reviewed, 25 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 196 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3X2q4ir1LOmf9Tep:-NPmtdGN1rSoAQaJiiI4:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L196)):*

Yeah just `return True` here should do fine.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 199 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3OrX8yh7nkmewtbc:-NPmvQBT6FIi1MyZCC1Y:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L199)):*

No, just passing would be a memory leak since the callback would still be stored by the publisher. The current implementation is correct.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 202 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3QqWBZ4xJCO0gZMC:-NPmvwNu77-0CptF1QY2:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L202)):*

No, if prompt_for_properties returns a false-y value then the dashboard cancels widget creation. The current implementation is correct.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 26 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQtDC1kZ3WT2gbvhR:-NPms53fCRx6wPDWs2_f:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L26)):*

NUM_PULSES
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 28 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTQqP34KvyzxdG0bbo:-NPms79OAL7xvQwgQdzO:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L28)):*

Constants should be ALL_CAPS_CASE. Probably "MAX_MESSAGE_BYTES" is a better name than NUMBER_OF_BYTES.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 32 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTRsRS8V-GOsybpzNw:-NPmsCK5EvBvyWocPnrL:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L32)):*

Props field is unnecessary like Kavin mentioned below.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 43 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTStcu5gvU7qk3hlzt:-NPmsnew8-V_3j4XLxy8:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L43)):*

I would rename `setup` -> `create` and `logicfy` -> `setup`. Also `snake_case` for function names please.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 106 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTTmIF5KAYHFeEOvOB:-NPmsn-l5WdvLZKqEATS:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L106)):*

`pulse_widgets` doesn't make sense as a name in this context. Probably best to rename to something like `check_message_incomplete`.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 146 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTX21x88w1rpV6DCd9:-NPmslT0-BrrDb8fGK1i:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L146)):*

`get_invalid_inputs` would be better since it includes the message type field.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 153 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTWIUY47lkz8mUsvxu:-NPmsv30DmrGSfZ5B2-b:byciooh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sinks/dashboard/items/can_sender/can_sender.py#L153)):*

If I try to type in something invalid like "hi" into one of the data byte inputs, will this catch that? I see that there is the QT regex validator above but will that let me type in something like that and just be mad about it or will it actually prevent me from typing invalid stuff in?
</blockquote></details>

The regex validator prevents you from typing invalid characters

itslinotlie (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95)* status: all files reviewed, 25 unresolved discussions (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 126 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG4MXc37WFQ_9ULX0C:-NPn0df8A_bm7bVO_dEB:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L126)):*

The error here is because the CAN parser (very end of `dashboard/parsers.py`) listens to the "CAN" channel. The fix here is to just change it to only listen to "CAN/Parsley" channel since it evidently depends on Parsley's formatting.

Also just noting that we should let the user control the timestamp bytes so they should be zero-initialized, not with `time.time()`.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_sender/can_sender.py` line 199 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPG3OrX8yh7nkmewtbc:-NPn0tBu-N2A6fdF4sHg:bubnv11) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/72b9eedb38f4da83c575689313cd4de3b7540b28/sinks/dashboard/items/can_sender/can_sender.py#L199)):*

Done.
</blockquote></details>

I don't understand how reviewable works, but anyways conclusion: since we don't subscribe to anything, there's nothing to unsubscribe and thus, passing will do
___
*[`sources/parsley/main.py` line 6 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NPTXTRTERsqkTsUmVIi:-NPn147q9_Oo14fpkF6z:b-kb1afm) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2fa168d0f83c9cb1a9de0149d089e843f1aaee1e/sources/parsley/main.py#L6)):*

You should be able to just `import message_types as mt`.
</blockquote></details>

# maybe its a my end problem, but ```import message_types as mt``` doesn't work
# this'll be replaced with an actual import with parsley rewrite so just ignore for now

BluCodeGH (changes requested):
Looks solid! PyQT wizardry.

> Reviewed 1 of 3 files at r4, 4 of 13 files at r5, 2 of 10 files at r6, 4 of 15 files at r7, 6 of 10 files at r8, 6 of 6 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-:-NUEeieb36HuraQVizUg:bma2l35)* status: all files reviewed, 21 unresolved discussions (waiting on @Darth-Raazi, @itslinotlie, and @KavinSatheeskumar)

___
*[`parsley` line 1 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NTdyMnK6rgw7k4shUid:-NTdyMnK6rgw7k4shUie:bnjiwip) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f9729ecfdbbeddb584f0fb02a624506b1b74c642/parsley#L1)):*
> ```
> Subproject commit 53c69673d56141d0f748c0a4a8bc7b57b80bd532
> ```

Reminder to change this when you merge parsley.
___
*[`sinks/dashboard/dashboard.py` line 27 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEVsv39fFdki2JdxCd:-NUEVsv39fFdki2JdxCe:b6o1f5f) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/dashboard.py#L27)):*
> ```Python
> from items.can_sender import CanSender
> from omnibus.util import TickCounter
> from utils import prompt_user, EventTracker
> ```

duplicated imports?
___
*[`sinks/dashboard/items/can_sender.py` line 21 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEXjgkA73dK8eyR0wc:-NUEXjgkA73dK8eyR0wd:builrpx) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L21)):*
> ```Python
> from .dashboard_item import DashboardItem
> from utils import EventTracker
> @Register
> ```

Blank line above this? Generally looks like you're in need of an autopep8
___
*[`sinks/dashboard/items/can_sender.py` line 95 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEYrP10zcngZjONT-5:-NUEYrP10zcngZjONT-6:b5jw7nw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L95)):*
> ```Python
>         self.channel = "CAN/Commands"
> 
>     def display_can_fields(self, fields: List[pf.Field]):
> ```

Please add one-line comments to each of the non-trivial functions in this class explaining what they do / when they are called.
___
*[`sinks/dashboard/items/can_sender.py` line 150 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEZ5YpCoT9d1VM2iUO:-NUEZ5YpCoT9d1VM2iUP:b-soojb0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L150)):*
> ```Python
>                 self.display_can_fields(nested_fields)
> 
>     def update_can_msg(self, text: str):
> ```

Probably rename to `update_switch`?
___
*[`sinks/dashboard/items/can_sender.py` line 182 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEZLzk0f-MuF-8UlpA:-NUEZLzk0f-MuF-8UlpB:b-a9c2n) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L182)):*
> ```Python
>             return
>         # if every field successfully encoded its data, long pulse all of the fields once
>         self.pulse_indexes = [i for i in range(0, self.widget_index)]
> ```

`list(range(self.widget_index))`
___
*[`sinks/dashboard/items/can_sender.py` line 189 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEZzbs30Qyh6Fptnjx:-NUEZzbs30Qyh6Fptnjy:b-fogce6) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L189)):*
> ```Python
>             "data": bit_data,
>             "length": bit_length
>         }
> ```

Hmm, I'm not a big fan of this format. Generally it's nice when the data flowing through omnibus is decently readable json rather than wrapped binary data. I'm not sure what the best approach here is but one option to consider is some sort of custom serialization for parlsey Fields, or perhaps being able to generate the dictionary format that parsley decodes from USB debug? More thought is needed, let's discuss in slack.
___
*[`sinks/dashboard/items/can_sender.py` line 193 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUE_EY02qk28INri0ul:-NUE_EY02qk28INri0um:bl99yc7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L193)):*
> ```Python
> 
>     def parse_can_msg(self) -> Union[BitString, None]:
>         self.pulse_indexes = []
> ```

Nit but plural of index is indices
___
*[`sinks/dashboard/items/can_sender.py` line 204 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUE_PPk4_h5SqSywcpj:-NUE_PPk4_h5SqSywcpk:b-3loc7i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L204)):*
> ```Python
>                     text = float(text)
>                 bit_str.push(*field.encode(text))
>             except Exception as e:
> ```

Can we get away with only catching `ValueError`s here? That would be preferable to a catch-all that would also catch coding errors on our part.
___
*[`sinks/dashboard/items/can_sender.py` line 205 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUE_YFb2nY9Lbpq3TYp:-NUE_YFb2nY9Lbpq3TYq:bgsy9ce) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L205)):*
> ```Python
>                 bit_str.push(*field.encode(text))
>             except Exception as e:
>                 print(f"{field.name} | {e}")
> ```

Unnecessary?
___
*[`sinks/dashboard/items/can_sender.py` line 216 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEa7Hw94cu8uX10lMb:-NUEa7Hw94cu8uX10lMc:b5yesmr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sinks/dashboard/items/can_sender.py#L216)):*
> ```Python
>     def pulse(self, pulse_invalid: bool):
>         self.pulse_count = 0
>         self.invalid = pulse_invalid
> ```

I think you can simplify this logic and remove the `self.invalid` flag by making `pulse_count` start at the desired number of pulses (times two) and count down.
___
*[`sources/parsley/main.py` line 30 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEahfBDbBgC9sHcq4q:-NUEahfBDbBgC9sHcq4r:b-qk61oa) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sources/parsley/main.py#L30)):*
> ```Python
>     parser.add_argument('baud', type=int, nargs='?', default=115200,
>                         help='the baud rate to use')
>     parser.add_argument('--format', default='telemetry',
> ```

While you're here can you update this default to `usb` please?
___
*[`sources/parsley/main.py` line 64 at r14](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NUEbGNECOC3lttv24NL:-NUEbGNECOC3lttv24NM:b-ppzcph) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/dc8c6b5ec6ac88d68793f3d55a0a473fc8238507/sources/parsley/main.py#L64)):*
> ```Python
> 
>         line = communicator.read()
>         if line is not None or line is input:
> ```

Hmm, it's going to be a massive pain to implement non-blocking input from stdin. Maybe we should just remove support for reading from stdin / writing to stdout and enforce passing the serial port. Tbh there isn't really a use case for that now that we have sending anyways.

BluCodeGH (changes requested):
> Reviewed 5 of 9 files at r16, 2 of 6 files at r22, 1 of 4 files at r23, 4 of 4 files at r24, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95)* status: all files reviewed, 11 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, @itslinotlie, and @KavinSatheeskumar)

___
*[`parsley` line 1 at r24](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NV7eCcW1YA8r0LJykqx:-NV7eCcW1YA8r0LJykqy:bkz4jf1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c271fa7d904ecd06b0fbc376d4e10af54badde3/parsley#L1)):*
> ```
> Subproject commit d82336746ce50b80c2c0fefd5ccd704b40bd7bd8
> ```

Note that this will need to change again when we merge parsley.
___
*[`sinks/dashboard/items/can_sender.py` line 190 at r24](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NV7dIcX37dLHWKK8kuf:-NV7dIcX37dLHWKK8kug:bam83vo) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c271fa7d904ecd06b0fbc376d4e10af54badde3/sinks/dashboard/items/can_sender.py#L190)):*
> ```Python
>     def send_can_message(self):
>         try:
>             bit_str = self.parse_can_msg()  # contains the message data bits
> ```

Reminder about our conversation to have parse_can_msg directly create and return the dictionary so we don't need to go through bitstring and back
___
*[`sources/parsley/main.py` line 55 at r24](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-NV7e1cq7lnvjrzhxtUx:-NV7e1cq7lnvjrzhxtUy:b-q4dkpa) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7c271fa7d904ecd06b0fbc376d4e10af54badde3/sources/parsley/main.py#L55)):*
> ```Python
>             if args.solo:
>                 print(formatted_string)
>             else:
> ```

We should always print.

BluCodeGH (approved):
:lgtm:

> Reviewed 4 of 4 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/95#-:-NVD1S_QA_HCjgMK7-3V:bnfp4nl)* status: all files reviewed, 9 unresolved discussions (waiting on @Darth-Raazi, @itslinotlie, and @KavinSatheeskumar)

PR #96, opened 2023-03-09
UI Restructure

Darth-Raazi:
Rewrote Dashboard to use the QGraphicsView framework, which uses a floating window style instead of the DockArea's grid style

# Kept Features
- Adding widgets
- Saving to file
- Loading from file
- Switching to a different file

# Removed features
- Deleting widgets via mouse
    - This would have required additional nesting of widgets
- Resizing widgets
    - I suspect this would also need additional nesting, not 100% sure
    - ~Didn't see the point now that we're not constrained to a grid but I can look into it if need be~
    - To be addressed in a separate PR

# New Features
- There is now ~an infinite~ a very large area underneath the view to allow for lots of scrolling, panning and zooming
    - Click and drag --> pan
    - Move mouse wheel (two fingers on trackpad) --> scroll
    - Hold shift while scrolling --> horizontal scroll
    - Hold control/cmd while doing scroll movement --> zoom
- Widgets are selectable and movable
- Pressing `Backspace` deletes a widget
- Key combinations
    - `Control =` --> zoom in (`Control +` but you don't have to press `Shift`)
    - `Control -` --> zoom out
    - `Control 0` --> reset view and zoom to center
- Lock mode:
    - Enable/disable via menu bar
    - Disables moving, deleting and adding widgets
    - Disables saving, loading and switching layout to/from file
    - Shows a `LOCKED` message in the window title

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96)


itslinotlie (commented):
In terms of quality of life:
- it would be dope if trackpad pinch gestures to zoom in/out was supported 
- unable to scroll/pan/zoom when the mouse is on top of a dashboard item
  - 
  - I've noticed these console messages whenever the mouse is on top of a dashboard item
- having a keyboard shortcut (i.e. cntrl/cmd 0) to reset the dashboard view to the center


In the 30minutes of playing around with the code, I've noticed a couple of bugs / inconveniences (that I haven't had the time to pinpoint yet)
- when pressing shift to initiate the zooming, occasionally I'll get sent to the top-left corner 
- when I zoom in/out too many times, the zoom no longer has the same "interactivity" as the initial zooming (it gets progressively  worse)
  => there should probably be a hard cap to how much you can realistically zoom out (I think when you try zooming out / zooming in too many times, the boundaries gets messed up)


For reference, I've been testing this on a mac (not sure how much of a difference this makes) and it feels janky.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQTUqtB7Lh-2QzSLRc_:btz6c6l)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NQTRB41-_TJ5W7j2-Ty:b-wjh8tx) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*
> ```Python
>     def wheelEvent(self, event):
>         # Zoom if Shift is held, otherwise scroll
>         if event.modifiers() & Qt.ShiftModifier:
> ```

I think Qt.ControlModifier would be more intuitive for zooming in/out

Darth-Raazi (commented):
Thanks for pointing these out. It was working pretty well on my machine, but I tried another laptop and noticed a lot of the same issues you brought up, so I'll have to look into those.

1. Qt is kind of bad with trackpads, I didn't see any handlers for pinching (and other multi-touch trackpad gestures) but I'll check again
2. Yeah I fought with this for a while, it has to do with what item is sent the mouse event. I'm not sure if I can get the widget to pass the event on to the view for scroll/zoom. Panning won't work on top of a widget by design because that's how you move a widget around
3. Getting the same messages, I looked it up and those seem to be because Qt can't properly handle Mac trackpads. Not much I can do there unfortunately (other than maybe suppress them?)
4. Keyboard shortcut for centring is a good idea, I'll implement that
5. Noticed this on the second laptop as well, no idea why that's happening but I'll take a look
6. I haven't run into this issue on either machine, but a zoom limit may be a good idea

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQWMphcD0VWGAKtFXrN:b25lqst)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NQWPa912Fvk1OzmXosZ:b-iefkyi) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*

I think Qt.ControlModifier would be more intuitive for zooming in/out
</blockquote></details>

Shift is a bit easier to quickly reach, especially on laptop and Mac keyboards

itslinotlie (changes requested):
I'm not sure if I follow your 2nd point (the "by design" part).

I guess what I wanted to say was that during testing, there were times when I would zoom in/out and a graph would take up the entire python application. And then because you can't zoom in/out or pan when the mouse is on top of a widget (which would be the case since the widget is the entire screen), you're kinda stuck there. (sometimes I'm able to zoom out but zooming has been very finicky for me so that hasn't always been the case)


I think being able to address panning over widgets would be pretty huge (I thought it was a bug when I first encountered that)

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQWzsw0En--QBNPQMyg:b4rhgjj)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NQWxZmm0l0uvN0EF231:b-rswzan) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

Shift is a bit easier to quickly reach, especially on laptop and Mac keyboards
</blockquote></details>

I think personal preferences shouldn't influence design decisions (at least for the default options). If you think about it, you're basically redefining the omnibus user experience and it doesn't make sense to have people refer to a manual in order to navigate around. It makes sense if the commands are standardized with what people are already familiar with.

we can ask around, but I think in terms of standards:
control + scrolling = zooming in/out => and then for mac its cmd instead of control
scrolling = up/down movement
shift + scrolling = left/right movement

I think if you really have a strong preference towards shift because its easier to reach, you can consider allowing user customization for these actions (and storing them in a file or something so that you can access them across sessions)

KavinSatheeskumar (changes requested):
Hello, my thoughts

* Firstly, as a windows user with a mouse, I have not been facing the issue Michael has described (probably because its all due to Mac track pads).
* I have noticed one bug, I can't zoom out :stuck_out_tongue_winking_eye: , I found the cause and left a comment as to the fix. Its a one character issue
* It is kind of annoying that I cannot resize widgets. Especially when we add a P&ID to the dashboard, that will be an issue. I would say that is close to comp critical, and we should have it for the 15th, but if its too much work, dw about it for this PR.
   * As a lesser goal, we could have the size saved in the config file, so that I could edit it before hand and get it nice and sized correctly, even if we don't provide UI support for re-sizing things.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQhXvGa6ebQFJjBcQr3:bxzryd6)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @itslinotlie)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NQhXBMZ3zutEwgABqEE:bpp7q1k) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*

I think personal preferences shouldn't influence design decisions (at least for the default options). If you think about it, you're basically redefining the omnibus user experience and it doesn't make sense to have people refer to a manual in order to navigate around. It makes sense if the commands are standardized with what people are already familiar with.

we can ask around, but I think in terms of standards:
control + scrolling = zooming in/out => and then for mac its cmd instead of control
scrolling = up/down movement
shift + scrolling = left/right movement

I think if you really have a strong preference towards shift because its easier to reach, you can consider allowing user customization for these actions (and storing them in a file or something so that you can access them across sessions)
</blockquote></details>

Adding customization for this this small is probably not worth it in the long run. I am ambivalent to which is used, although I will note that 95% of the laptops running this code will be using windows.
___
*[`sinks/dashboard/dashboard.py` line 48 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSIJiX9a4SyMwvrrZV:-NQSIJiX9a4SyMwvrrZW:b-f4dgpf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L48)):*
> ```Python
>             if angle.x() > 0 or angle.y() > 0:
>                 zoomFactor = zoomInFactor
>             elif angle.x() < 0 or angle.y() > 0:
> ```

I think it should be as below. (Direction of second inequality on this line flipped)



_Code snippet:_
```Python
elif angle.x() < 0 or angle.y() < 0:
```
___
*[`sinks/dashboard/dashboard.py` line 95 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSTee63nCu9bmNgRbe:-NQSTee63nCu9bmNgRbf:b-p3pkfo) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L95)):*
> ```Python
>         # List to keep track of menu bar action that
>         # can be disabled when dashboard is locked
>         self.disable = []
> ```

Can we rename this to self.actions? rather than disable. Just more indicative of what it is + it seems like you are doing some event handler stuff with the name "disable"
___
*[`sinks/dashboard/dashboard.py` line 162 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSNmeA7BKbaIRFytp2:-NQSNmeA7BKbaIRFytp3:bs719xh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L162)):*
> ```Python
> 
>     # Method to add widgets
>     def add(self, dashitem, pos=None):
> ```

Can we make the widget re-sizable? If its too much work, we can make that a future PR

BluCodeGH:
Just noting that we should leave resizing entirely out of the scope of this PR and address it in future ones. While valuable it's by no means critical and I want to get this merged to unblock parallel work.

Darth-Raazi (commented):
The only thing left is to enable scrolling/panning while the mouse is over a widget. I honestly have no idea why this isn't working, the mouse events are even being forwarded to the proper function. Might look into it more with a fresh mind later.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQn5anR0tPkBShlM-bx:bs4dcso)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NQn4fM92PsvPeoreNEs:b-tnlen7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*

Adding customization for this this small is probably not worth it in the long run. I am ambivalent to which is used, although I will note that 95% of the laptops running this code will be using windows.
</blockquote></details>

Keeping this as shift to zoom for now to simplify some of the event handling code. Control is used in some other key combinations.
___
*[`sinks/dashboard/dashboard.py` line 48 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSIJiX9a4SyMwvrrZV:-NQn5Ljg8wuQ7jT9OL3V:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L48)):*

I think it should be as below. (Direction of second inequality on this line flipped)
</blockquote></details>

Done
___
*[`sinks/dashboard/dashboard.py` line 95 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSTee63nCu9bmNgRbe:-NQn5OxOBb-Ka6uI8lu_:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L95)):*

Can we rename this to self.actions? rather than disable. Just more indicative of what it is + it seems like you are doing some event handler stuff with the name "disable"
</blockquote></details>

Done
___
*[`sinks/dashboard/dashboard.py` line 162 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSNmeA7BKbaIRFytp2:-NQn5TKm3SW2lXCcawXy:b-bx6azq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L162)):*

Can we make the widget re-sizable? If its too much work, we can make that a future PR
</blockquote></details>

Separate PR

Darth-Raazi (commented):
Figured it out, scrolling while your mouse is over a widget works now

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQvTeX-9diZ9mMyIucx:bhv513v)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

itslinotlie (dismissed):
Besides the keyboard shortcuts (but seeing how everyone is fine with it, then whatever), you'll need to
- update the omnibus README to mention that ```Python 3.10 or newer is required``` with the addition of your ```match ... case ...```
- preferably silence the dispatch logs in the terminal
- 

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NQwFH-I2PLlxvEXPKt0:b-oy7rtj)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 368 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR1GEvr11hZ_ONWv6fj:-NR1GEvr11hZ_ONWv6fk:b-95xzrd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L368)):*
> ```Python
>                     self.view.zoom("in")
>                 case Qt.Key_Minus:
>                     self.view.zoom("out")
> ```

These keyboard shortcuts are a great addition. I think the scaling factor for these keyboard commands can be more sensitive though (somewhere around 1.5 feels nice)

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @itslinotlie)

___
*[`sinks/dashboard/dashboard.py` line 240 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR5iDBP1C9dxklpEn5h:-NR5iDBP1C9dxklpEn5i:bp1annz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L240)):*
> ```Python
> 
>         # Then load the data from the savefile
>         if os.path.exists(self.filename):
> ```

This is pedantic AF, but I personally prefer the "gaurd claus" heuristic for things like this. (Code snippit below). It saves one level of indentation.



_Code snippet:_
```Python
if !os.path.exists(self.filename):
    return
    
with open(self.filename, "r") as savefile:
    data = json.load(savefile)
```

BluCodeGH (changes requested):
An issue I noticed: if I save the dashboard, zoom in, and then load from file, the widgets get placed such that they overlap. The loading should be independent of the current zoom level.

> Reviewed 2 of 2 files at r3, all commit messages.
> Dismissed @KavinSatheeskumar and @itslinotlie from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NR8QFt085jowBlkQwsU:b-pcs2jz)* status: all files reviewed, 8 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 95 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSTee63nCu9bmNgRbe:-NR8Si6nBZGOcSmbLSu3:brirt13) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L95)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

Done
</blockquote></details>

Sorry to make you rename again but I think `lockableActions` is a much clearer name.
___
*[`sinks/dashboard/dashboard.py` line 34 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8QdEz3fg1H9S8sp0D:-NR8QdEz3fg1H9S8sp0E:bik9tw0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L34)):*
> ```Python
>         self.zoomed = 1
> 
>     def zoom(self, direction: str):
> ```

Using a string parameter here is kinda sus. I think the value of a `zoom` method is that we can have some global 'sensitivity' parameter, and then we take a float argument, multiply it by the sensitivity, and call `self.scale` with that.
___
*[`sinks/dashboard/dashboard.py` line 40 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RgXH7kTrI2p3-t8c:-NR8RgXH7kTrI2p3-t8d:b-scwcbv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L40)):*
> ```Python
> 
>         # Zooms to the position of the mouse
>         self.setTransformationAnchor(QGraphicsView.AnchorUnderMouse)
> ```

This call can be moved to `__init__` and only called once.
___
*[`sinks/dashboard/dashboard.py` line 53 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RklJ36GoyBsrhAPx:-NR8RklJ36GoyBsrhAPy:btkofx7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L53)):*
> ```Python
>     def wheelEvent(self, event):
>         # Zoom if Shift is held, otherwise scroll
>         if event.modifiers() == Qt.ShiftModifier:
> ```

Changing this to ControlModifier doesn't break anything and makes the dashboard significantly more intuitive to use. What I would do here is let either shift or control zoom.
___
*[`sinks/dashboard/dashboard.py` line 57 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RuuO3mUZ6sxsPEnU:-NR8RuuO3mUZ6sxsPEnV:bo686s3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L57)):*
> ```Python
>             angle = event.angleDelta()
>             if angle.x() > 0 or angle.y() > 0:
>                 self.zoom("in")
> ```

On high-resolution trackpads / scroll wheels (which send 8x the mouse events each of 1/8 the magnitude) this zooms way too fast. You should change zoom to take a float and use the actual wheel event angle as a scalar for how much to zoom.
___
*[`sinks/dashboard/dashboard.py` line 178 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8Tcr06LYAwZweFINa:-NR8Tcr06LYAwZweFINb:b-1faz9y) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L178)):*
> ```Python
>         # the widget in the center of the current view
>         if pos:
>             mapped = self.view.mapToScene(pos[0], pos[1])
> ```

I think this is the source of your save/load issues. You should be using scene coordinates not view coordinates in the savefile, and likewise specify scene coordinates to add(). If you don't specify coordinates then using the center of the view is good.

Darth-Raazi (commented):
Fixed

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRUZSkz7gu97LzDVtQE:b-886gsx)* status: 1 of 3 files reviewed, 8 unresolved discussions (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 95 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQSTee63nCu9bmNgRbe:-NRPHaGX0RQjyJVfHUMC:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L95)):*

Sorry to make you rename again but I think `lockableActions` is a much clearer name.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 34 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8QdEz3fg1H9S8sp0D:-NRUdBVWEh7zjEXddphV:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L34)):*

Using a string parameter here is kinda sus. I think the value of a `zoom` method is that we can have some global 'sensitivity' parameter, and then we take a float argument, multiply it by the sensitivity, and call `self.scale` with that.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 40 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RgXH7kTrI2p3-t8c:-NRPH48YEjSnr1x868Hr:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L40)):*

This call can be moved to `__init__` and only called once.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 53 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RklJ36GoyBsrhAPx:-NRPHIKr3NVSaGTbIe4G:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L53)):*

Changing this to ControlModifier doesn't break anything and makes the dashboard significantly more intuitive to use. What I would do here is let either shift or control zoom.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 57 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8RuuO3mUZ6sxsPEnU:-NRUZ2nB8zphl0VroM1O:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L57)):*

On high-resolution trackpads / scroll wheels (which send 8x the mouse events each of 1/8 the magnitude) this zooms way too fast. You should change zoom to take a float and use the actual wheel event angle as a scalar for how much to zoom.
</blockquote></details>

Done
___
*[`sinks/dashboard/dashboard.py` line 178 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8Tcr06LYAwZweFINa:-NRUZB_o92lCzSmbSKX0:b-25254w) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L178)):*

I think this is the source of your save/load issues. You should be using scene coordinates not view coordinates in the savefile, and likewise specify scene coordinates to add(). If you don't specify coordinates then using the center of the view is good.
</blockquote></details>

It was actually because I was adding the widgets before zooming when I loaded. Fixed that.
___
*[`sinks/dashboard/dashboard.py` line 240 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR5iDBP1C9dxklpEn5h:-NRPI4NP7TcAevZwCm2n:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L240)):*

This is pedantic AF, but I personally prefer the "gaurd claus" heuristic for things like this. (Code snippit below). It saves one level of indentation.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 368 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR1GEvr11hZ_ONWv6fj:-NRUZKuZD74rXy22NZV-:bfjg9nn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L368)):*

These keyboard shortcuts are a great addition. I think the scaling factor for these keyboard commands can be more sensitive though (somewhere around 1.5 feels nice)
</blockquote></details>

Changed to use a float value instead

BluCodeGH (approved):
A couple nitpicky comments that you should only fix if other reviewers leave more important comments.

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRV6aRgAktRqnY3zsgD:b-pfw6fa)* status: all files reviewed, 5 unresolved discussions (waiting on @Darth-Raazi, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 178 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8Tcr06LYAwZweFINa:-NRV4ZeI7X4c18SzQIJR:b-n0vnso) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L178)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

It was actually because I was adding the widgets before zooming when I loaded. Fixed that.
</blockquote></details>

Architecturally I think using view coordinates to save doesn't make a lot of sense when we have scene coordinates that make all these problems go away, but I'd like to get this in so I won't block on it.
___
*[`sinks/dashboard/dashboard.py` line 57 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NRV5OhNDtNzMPW78pOI:-NRV3iMD0X-IikTs_Sw3:b-yx8xj0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/844c510999e2c373cd89637f018df3e959e3310b/sinks/dashboard/dashboard.py#L57)):*
> ```Python
>                 self.zoom(float(angle.x()))
>             else:
>                 self.zoom(float(angle.y()))
> ```

`self.zoom(angle.manhattanLength())` works in the general case. Note that the cast from int to float is unnecessary, if you want to satisfy a typechecker I'd just change the type of `self.zoom` since we never call with floats anyways.
___
*[`sinks/dashboard/dashboard.py` line 250 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NRV527h6gJJHxITWyhi:-NRV527iE6O5ps9ZEXGy:b-v19wlt) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/844c510999e2c373cd89637f018df3e959e3310b/sinks/dashboard/dashboard.py#L250)):*
> ```Python
>         new_zoom = data["zoom"]
>         self.view.scale(new_zoom, new_zoom)
>         self.view.zoomed = new_zoom
> ```

nit but breaking out these values into new variables (`curr_zoom`, `new_zoom`) makes the code harder to read.

KavinSatheeskumar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRVU4eQ7IbnJvbRn81s:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @Darth-Raazi and @itslinotlie)

itslinotlie (commented):
I still stand by what I said before: I think having an intuitive and easy-to-pick-up interface is extremely important (especially now that a whole bunch of people are going to be working on Omnibus), which is currently lacking to some degree. If it's not addressed in this PR, I would like to see it addressed in a future PR.

> Reviewed 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRVUg1A9zAiMLYsm38H:bjhpx45)* status: all files reviewed, 4 unresolved discussions (waiting on @Darth-Raazi and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NRVT9YV2cYAtojJrPfR:brbmsvg) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

Keeping this as shift to zoom for now to simplify some of the event handling code. Control is used in some other key combinations.
</blockquote></details>

I think I might have emphasized the wrong points earlier. To my understanding, it is currently not possible to navigate around Omnibus with just a mouse and keyboard. You can only scroll up and down, but there is no support for scrolling left-right with a mouse.

My proposal was to follow the convention that other platforms generally follow
- holding shift + scrolling = sideways scrolling
- holding control + scrolling = zooming in and out

I was wrong about the cmd thing, turns out the control equivalent for Mac is command, so I don't think you have to worry about windows/mac in terms of the control key.
___
*[`sinks/dashboard/dashboard.py` line 368 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR1GEvr11hZ_ONWv6fj:-NRVUKp8E6fOo21ZNJeu:b-kp2db5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L368)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

Changed to use a float value instead
</blockquote></details>

I don't see how changing the type to float changes anything functionally. My point was that zooming in and out with the keyboard is too insensitive

Darth-Raazi (commented):
Added everything you wanted plus a help box that lists navigation commands

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRoK8guDwQ_cMctRyMR:b9v9ev)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NRoBSTE0fcfQTi_H0iz:b-trcvr9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*

I think I might have emphasized the wrong points earlier. To my understanding, it is currently not possible to navigate around Omnibus with just a mouse and keyboard. You can only scroll up and down, but there is no support for scrolling left-right with a mouse.

My proposal was to follow the convention that other platforms generally follow
- holding shift + scrolling = sideways scrolling
- holding control + scrolling = zooming in and out

I was wrong about the cmd thing, turns out the control equivalent for Mac is command, so I don't think you have to worry about windows/mac in terms of the control key.
</blockquote></details>

I have looked a lot and can't seem to find a good solution for horizontal scrolling. I've implemented one for your satisfaction, but it's choppy and scrolls way too fast. The other implementation that comes to mind is to capture the event, create an entirely new one, copy over every attribute save for the angleDelta(), construct a new QPoint with the values from angleDelta() but swapped, assign that as the new angleDelta() and then forward the new event to the super class. Complicated and I'm not even sure if it will work properly, so perhaps this can be addressed in a future PR. 

If you're using just a mouse, I suggest click and drag on the view to move horizontally for now. Control/CMD + scroll zooms, regular scrolling moves up and down.
___
*[`sinks/dashboard/dashboard.py` line 178 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR8Tcr06LYAwZweFINa:-NRe2D1IA3SAQwhHsyde:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L178)):*

Architecturally I think using view coordinates to save doesn't make a lot of sense when we have scene coordinates that make all these problems go away, but I'd like to get this in so I won't block on it.
</blockquote></details>

Done
___
*[`sinks/dashboard/dashboard.py` line 368 at r3](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NR1GEvr11hZ_ONWv6fj:-NRo2CWE9bqLT2-AKq5h:booy1j5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/2dc2c4f4462235f5ba3e7792f7be2f9d90bcd971/sinks/dashboard/dashboard.py#L368)):*

I don't see how changing the type to float changes anything functionally. My point was that zooming in and out with the keyboard is too insensitive
</blockquote></details>

Changed the sensitivity to 200. Feels good to me.
___
*[`sinks/dashboard/dashboard.py` line 57 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NRV5OhNDtNzMPW78pOI:-NRo21U6ABqz_zeRBvUq:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/844c510999e2c373cd89637f018df3e959e3310b/sinks/dashboard/dashboard.py#L57)):*

`self.zoom(angle.manhattanLength())` works in the general case. Note that the cast from int to float is unnecessary, if you want to satisfy a typechecker I'd just change the type of `self.zoom` since we never call with floats anyways.
</blockquote></details>

Done
___
*[`sinks/dashboard/dashboard.py` line 250 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NRV527h6gJJHxITWyhi:-NRe38RR9PduM0TDYOyn:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/844c510999e2c373cd89637f018df3e959e3310b/sinks/dashboard/dashboard.py#L250)):*

nit but breaking out these values into new variables (`curr_zoom`, `new_zoom`) makes the code harder to read.
</blockquote></details>

Done

Darth-Raazi:
Also there's a bug that if, as soon as the dashboard loads, you try to zoom, you'll be sent to one of the corners. No idea what's causing it, just move your cursor slightly before zooming and it won't happen

itslinotlie (approved):
:lgtm:

> Reviewed 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRp0BXT9fJshIYlWDPv:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 36 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-NQTRB41-_TJ5W7j2-Tx:-NRp05_M7-wnMKKHFuac:bt2vr1b) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38973aa38d14e8c312e2be42805bb7784bee9efb/sinks/dashboard/dashboard.py#L36)):*
<details><summary><i>Previously, Darth-Raazi wrote…</i></summary><blockquote>

I have looked a lot and can't seem to find a good solution for horizontal scrolling. I've implemented one for your satisfaction, but it's choppy and scrolls way too fast. The other implementation that comes to mind is to capture the event, create an entirely new one, copy over every attribute save for the angleDelta(), construct a new QPoint with the values from angleDelta() but swapped, assign that as the new angleDelta() and then forward the new event to the super class. Complicated and I'm not even sure if it will work properly, so perhaps this can be addressed in a future PR. 

If you're using just a mouse, I suggest click and drag on the view to move horizontally for now. Control/CMD + scroll zooms, regular scrolling moves up and down.
</blockquote></details>

Fixed.

KavinSatheeskumar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/96#-:-NRpHZoy4CJiEkfAyrSX:bnfp4nl)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @itslinotlie)

PR #104, opened 2023-04-03
Added parameters to the new UI

AvKat:
Major changes:

1.  Added UI props (now parameters) to the existing UI components using pyqtgraph's ParameterTree
2. Save the width, height and position along with params, making it easy to load
3. Entirely got rid of the `prompt_user` functions. Instead, items are added with some default params and can easly be changed with a single click.

To add new items, refer to `dashboard_item.py`, where Jack has given an overview of ParameterTree and how to extend it. The only required functions for a new DashItem are `__init__` and `get_name` (static)

Limitations:
I could not figure out how to make Qt checkbox text wider so the stream names look like this:
Screenshot 2023-04-03 at 6 47 49 PM
Let me know if you have any ideas on how to fix this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104)


itslinotlie (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @AvKat, @BluCodeGH, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 183 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8sMZ0AmSN4YGVooYh:-NS8sMZ12ecokBt9NHi5:b4ztdck) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e369195c664e65bb684cb07976488dabf1664f15/sinks/dashboard/dashboard.py#L183)):*
> ```Python
>             self.parameter_tree.hide()
>             self.layout.removeWidget(self.parameter_tree)
>             self.parameter_tree = None
> ```

Is there a reason why we're destroying and creating a new parameter tree every time an item is selected? I tried the following and it seemed fine to me:

```
self.parameter_tree = ParameterTree(showHeader=False)

def onSelectionChanged(self):
  if len(items) != 1:
    self.parameter_tree.hide()
    return
  
  item = self.widgets[items[0]][1]
  self.parameter_tree.setParameters(item.get_parameters(), showTop=False)
  self.parameter_tree.show()
```
___
*[`sinks/dashboard/dashboard.py` line 189 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8orce54lHzhjyb-QX:-NS8orce54lHzhjyb-QY:bhcy3f3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e369195c664e65bb684cb07976488dabf1664f15/sinks/dashboard/dashboard.py#L189)):*
> ```Python
>         # Show the tree
>         item = self.widgets[items[0]][1]
>         self.parameter_tree = ParameterTree(showHeader=False)
> ```

I've noticed that when you modify a parameter in ```ParameterTree```, you aren't updating the underlying bounding box for said item. You're aware of ```sigTreeStateChanged``` so might want to update the bounding box in the dashboard file as well.

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104)* status: 0 of 7 files reviewed, 10 unresolved discussions (waiting on @AvKat and @BluCodeGH)

___
*[`sinks/dashboard/dashboard.py` line 87 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ogb1DaByO9nOrHH3:-NS8ogb1DaByO9nOrHH4:bf7xsi8) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L87)):*
> ```Python
> 
>         # The file from which the dashboard is loaded
>         self.filename = "savefile.json"
> ```

Not the best idea to change up how the whole file prompt system works rn, We can do this in another PR.
___
*[`sinks/dashboard/dashboard.py` line 100 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8oq4uDluYDnUAQNyi:-NS8oq4uDluYDnUAQNyj:bolfgbu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L100)):*
> ```Python
> 
>         # Create a grid layout
>         self.layout = QHBoxLayout()
> ```

Any reason for this switch? (Also, we should really fix that comment lmao
___
*[`sinks/dashboard/dashboard.py` line 118 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8pZ2iDsvcgx73VDO_:-NS8pZ2iDsvcgx73VDOa:bdm1tsc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L118)):*
> ```Python
>         # because of the way python handles
>         def create_registry_trigger(i):
>             def return_fun():
> ```

So this change here makes it impossible to add multiple plots at once, which kinda defeats the purpose of an older PR. I think this is subideal because now the work flow looks more like "Add an empty widget, change its props, add another, change its props" and so on. This is worse than just going through the prompt one time just selecting "plot separately".

I think the real solution here is to keep prompt for properties around for now. Maybe we can scale back the level of prompting that takes place during prompt for properties, but we shouldn't get rid of it entirely.
___
*[`sinks/dashboard/dashboard.py` line 204 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8rDAqBnfkFqvQD_l5:-NS8rDAqBnfkFqvQD_l6:b-mdlh9m) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L204)):*
> ```Python
>         # If a position is given, use it. If not, position
>         # the widget in the center of the current view
>         if pos:
> ```

I am guess that this diff here is due to the lack of a rebase. We can git de-screw later.
___
*[`sinks/dashboard/dashboard.py` line 204 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8s6g62q1y0QCfwp5q:-NS8s6g62q1y0QCfwp5r:bmgh6ro) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7805c542103df2e72b4596e425c7dacbf273dc20/sinks/dashboard/dashboard.py#L204)):*
> ```Python
>         # Create a rectangle around the proxy widget
>         # to make it movable and selectable
>         rect = QGraphicsRectItem(pos[0]-1, pos[1]-1, width+1, height+1)
> ```

When we add an item to the dashboard, the height and width here should be taken from the saved properties. (and ideally should be changed when the properties are updated.

Height and Width should be treated similarly to how pos is treated. 
___
*[`sinks/dashboard/utils.py` line 2 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8vFZC9gGmAU6UHHyB:-NS8vFZC9gGmAU6UHHyC:b-qy75v3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7805c542103df2e72b4596e425c7dacbf273dc20/sinks/dashboard/utils.py#L2)):*
> ```Python
> from pyqtgraph.Qt import QtWidgets
> from pyqtgraph.Qt.QtCore import Qt
> ```

As mentioned above, I think we should keep these for the reasons outlined above.
___
*[`sinks/dashboard/items/can_message_table.py` line 152 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ugOeEhoxNfjERR0k:-NS8ugOeEhoxNfjERR0l:bzbwncf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/can_message_table.py#L152)):*
> ```Python
>         #    is rendered in the right object
> 
>         if params:
> ```

As mentioned above, we should not be saving width and height as props, rather, we should save the height and width the same way position is saved.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 77 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8uOtd9-0EeqpFTtzc:-NS8uOtd9-0EeqpFTtzd:bnktotn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/plot_dash_item.py#L77)):*
> ```Python
>         self.layout.addWidget(self.widget, 0, 0)
> 
>     def on_series_change(self, param, value):
> ```

Instead of forcing every dash item to have an "on_props_change" function, why not just delete and re-create the dashitem everytime we change its props?



_Code quote (from sinks/dashboard/items/dashboard_item.py):_
```Python
get_serialized_parameters
```

itslinotlie (commented):
Regarding the ParameterTree column size, I've played around with the following which seemed alright:

``` python

from pyqtgraph.Qt.QtWidgets import QHeaderView
        self.parameter_tree = ParameterTree(showHeader=True)
        self.parameter_tree.setParameters(item.get_parameters(), showTop=True)
        header = self.parameter_tree.header()
        header.setMinimumSectionSize(100) # maybe this is bad for small window sizes, but it's a decent start
        header.setSectionResizeMode(QHeaderView.Interactive) # and then if you want more flexibility with the column size, this is that option

```

I think it would be nice if you could resize the parameter tree (so dragging the left side of the parameter tree instead of expanding the application window), but I tried looking into QSplitter and honestly I couldn't get anything to work in 30 minutes so I called it quits xd

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-:-NS8wnKx-62fE_WwA7OJ:b-ongjv3)* status: 0 of 7 files reviewed, 10 unresolved discussions (waiting on @AvKat and @BluCodeGH)

BluCodeGH (changes requested):
Looks like the rebase was done onto a different branch and so isn't reflected in this PR? Also remember to open reviewable and reply to people's comments with "Done" once you have addressed them.

> Reviewed 4 of 7 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-:-NSHVIib-uMmqeJLXaLJ:bgie8w7)* status: all files reviewed, 11 unresolved discussions (waiting on @AvKat, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 183 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8sMZ0AmSN4YGVooYh:-NSHRpXkDAx9YUQRpgs-:bv0ltcc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e369195c664e65bb684cb07976488dabf1664f15/sinks/dashboard/dashboard.py#L183)):*

Is there a reason why we're destroying and creating a new parameter tree every time an item is selected? I tried the following and it seemed fine to me:

```
self.parameter_tree = ParameterTree(showHeader=False)

def onSelectionChanged(self):
  if len(items) != 1:
    self.parameter_tree.hide()
    return
  
  item = self.widgets[items[0]][1]
  self.parameter_tree.setParameters(item.get_parameters(), showTop=False)
  self.parameter_tree.show()
```
</blockquote></details>

I think this was a relic of my prototyping and not realizing that `hide()` existed.
___
*[`sinks/dashboard/dashboard.py` line 87 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ogb1DaByO9nOrHH3:-NSHSZJrBD5gdrPnOAm0:b-v26jb4) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L87)):*

Not the best idea to change up how the whole file prompt system works rn, We can do this in another PR.
</blockquote></details>

I'd much rather keep the 5 line change that removes the last traces of `prompt_user` so we can get rid of it.
___
*[`sinks/dashboard/dashboard.py` line 118 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8pZ2iDsvcgx73VDO_:-NSHPRtz1Nr_UJn4U4Zf:bdhnaeb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L118)):*

So this change here makes it impossible to add multiple plots at once, which kinda defeats the purpose of an older PR. I think this is subideal because now the work flow looks more like "Add an empty widget, change its props, add another, change its props" and so on. This is worse than just going through the prompt one time just selecting "plot separately".

I think the real solution here is to keep prompt for properties around for now. Maybe we can scale back the level of prompting that takes place during prompt for properties, but we shouldn't get rid of it entirely.
</blockquote></details>

The whole `prompt_for_properties` architecture was really convoluted and messy (eg: string argument for prompt type in `prompt_user`), I think we should use this PR to tear it out and put in something significantly better.

The concern with any sort of bulk-add feature is how you lay out the plots on the dashboard. For any "real" dashboard being built for a test you're going to be laying things out by hand around a P&ID image, and generally won't have very many plots in the first place. So the main use case that bulk add serves is wanting to quickly show a bunch of sensors for stuff like DAQ testing.

I agree that this "automatic dashboard creating" for testing is really useful, but IMO trying to tie that functionality in with creating dashitems isn't necessary and leads to a much more complex architecture.

I think the way to do that is to have an entirely separate feature (that could be initiated from the menu bar) along the lines of "Automatic Dashboard" which prompts with a list of series to select from and then creates plots in a grid with all of those series. That work should probably start as soon as this PR goes in, but I'm comfortable temporarily not having the bulk add functionality in exchange for a much better architecture in the longer term.
___
*[`sinks/dashboard/utils.py` line 2 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8vFZC9gGmAU6UHHyB:-NSHSpMC3cdTPtJbCHhf:b-feff89) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7805c542103df2e72b4596e425c7dacbf273dc20/sinks/dashboard/utils.py#L2)):*

As mentioned above, I think we should keep these for the reasons outlined above.
</blockquote></details>

See my above longer explanation.
___
*[`sinks/dashboard/items/can_message_table.py` line 152 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ugOeEhoxNfjERR0k:-NSHTIW593Psk2FZN9hB:b-9ejtz0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/can_message_table.py#L152)):*

As mentioned above, we should not be saving width and height as props, rather, we should save the height and width the same way position is saved.
</blockquote></details>

Because height and width are parameters they should be saved/loaded like the rest of the parameters. However, it should be the superclass `DashboardItem.__init__` that handles loading them when saved, since it also handles all other operations regarding them.
___
*[`sinks/dashboard/items/dashboard_item.py` line 56 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NSHTfw86Nwe_4nrRv3H:-NSHTfw86Nwe_4nrRv3I:b-byvgll) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/016ab6078e5e59d70b593a70851209c92ecf7c4e/sinks/dashboard/items/dashboard_item.py#L56)):*
> ```Python
>         return self.parameters
> 
>     def get_serialized_parameters(self):
> ```

This logic should live in the dashboard save/load functions which just call to `get_parameters`. Currently this gives two redundant ways to get the parameters which doesn't make a ton of sense imo. Also remember that `parameter.saveState()` and `parameter.restoreState()` exist and probably work better for the save/load logic.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 77 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8uOtd9-0EeqpFTtzc:-NSHUGm3DZ0BqWqC96lH:b-igezr9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/plot_dash_item.py#L77)):*

Instead of forcing every dash item to have an "on_props_change" function, why not just delete and re-create the dashitem everytime we change its props?
</blockquote></details>

This is the more general solution (dashitems might want to do more than just be recreated on param change), I think we should leave it as is for now and once we have a better selection of dashitems we can decide whether the recreation approach is sufficient.

AvKat (commented):
Okay so the rebasing was becoming a bit of a pain so @KavinSatheeskumar asked me to move to a fresh branch (`new_unified_parameters`) and move all my changes there. I have addressed most of the things there and will be making a PR soon. The conflicting topics that are stoppig me are: 
1. Should we entirely get rid of `prompt_for_properties` or not? ( @BluCodeGH and @KavinSatheeskumar  need to discuss this )
2. Should we move the `pos` inside of parameters or not? (I need to discuss this with @KavinSatheeskumar )
3. Should we delete and recreate dashitem on parameter change? ( @BluCodeGH and @KavinSatheeskumar )

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-:-NSHfms15TyMzff1syr5:b-x9xwuj)* status: all files reviewed, 11 unresolved discussions (waiting on @BluCodeGH, @itslinotlie, and @KavinSatheeskumar)

___
*[`sinks/dashboard/dashboard.py` line 183 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8sMZ0AmSN4YGVooYh:-NSHe9_n3PH7_0bCbNKv:bkr0qqf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e369195c664e65bb684cb07976488dabf1664f15/sinks/dashboard/dashboard.py#L183)):*

I think this was a relic of my prototyping and not realizing that `hide()` existed.
</blockquote></details>

I have handled it in the new branch.
___
*[`sinks/dashboard/dashboard.py` line 189 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8orce54lHzhjyb-QX:-NSHeCoJ6baYHC2PSztB:b-9jn1cl) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/e369195c664e65bb684cb07976488dabf1664f15/sinks/dashboard/dashboard.py#L189)):*

I've noticed that when you modify a parameter in ```ParameterTree```, you aren't updating the underlying bounding box for said item. You're aware of ```sigTreeStateChanged``` so might want to update the bounding box in the dashboard file as well.



</blockquote></details>

Handled in the new branch.
___
*[`sinks/dashboard/dashboard.py` line 87 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ogb1DaByO9nOrHH3:-NSHga5cEAY1gWfv0QrJ:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L87)):*

I'd much rather keep the 5 line change that removes the last traces of `prompt_user` so we can get rid of it.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 100 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8oq4uDluYDnUAQNyi:-NSHdjETC6L0AlTozf5z:brwbbu9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L100)):*

Any reason for this switch? (Also, we should really fix that comment lmao
</blockquote></details>

Yeah this is because we want the ParamaterTree on the right and not bottom.
___
*[`sinks/dashboard/dashboard.py` line 118 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8pZ2iDsvcgx73VDO_:-NSHdvYa3nlMalgSvr9C:br939fn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L118)):*

The whole `prompt_for_properties` architecture was really convoluted and messy (eg: string argument for prompt type in `prompt_user`), I think we should use this PR to tear it out and put in something significantly better.

The concern with any sort of bulk-add feature is how you lay out the plots on the dashboard. For any "real" dashboard being built for a test you're going to be laying things out by hand around a P&ID image, and generally won't have very many plots in the first place. So the main use case that bulk add serves is wanting to quickly show a bunch of sensors for stuff like DAQ testing.

I agree that this "automatic dashboard creating" for testing is really useful, but IMO trying to tie that functionality in with creating dashitems isn't necessary and leads to a much more complex architecture.

I think the way to do that is to have an entirely separate feature (that could be initiated from the menu bar) along the lines of "Automatic Dashboard" which prompts with a list of series to select from and then creates plots in a grid with all of those series. That work should probably start as soon as this PR goes in, but I'm comfortable temporarily not having the bulk add functionality in exchange for a much better architecture in the longer term.
</blockquote></details>

Okay so can you guys discuss and come to a final decision on this so that I can add/remove it as required. (I have both ready to go)
___
*[`sinks/dashboard/dashboard.py` line 204 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8rDAqBnfkFqvQD_l5:-NSHeGhU-HdBQq7qWCTU:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/dashboard.py#L204)):*

I am guess that this diff here is due to the lack of a rebase. We can git de-screw later.
</blockquote></details>

Done.
___
*[`sinks/dashboard/dashboard.py` line 204 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8s6g62q1y0QCfwp5q:-NSHeJXG5_K_FEkUwe78:b-4vph64) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7805c542103df2e72b4596e425c7dacbf273dc20/sinks/dashboard/dashboard.py#L204)):*

When we add an item to the dashboard, the height and width here should be taken from the saved properties. (and ideally should be changed when the properties are updated.

Height and Width should be treated similarly to how pos is treated. 
</blockquote></details>

After the playing with it for a while, and our discussion yesterday, I feel like the current approach is the best one because as you said, we may wanna add drag-to-resize later, which will move width outside. I tried moving pos inside parameters but it is a a very heavy task and reversing it would also take time so I would recommend not going through with it rn.
___
*[`sinks/dashboard/utils.py` line 2 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8vFZC9gGmAU6UHHyB:-NSHew0E-a3mGjwCQFLO:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7805c542103df2e72b4596e425c7dacbf273dc20/sinks/dashboard/utils.py#L2)):*

See my above longer explanation.
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/can_message_table.py` line 152 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8ugOeEhoxNfjERR0k:-NSHeyiK9JZ1YGHtTbco:bg6xy9w) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/can_message_table.py#L152)):*

Because height and width are parameters they should be saved/loaded like the rest of the parameters. However, it should be the superclass `DashboardItem.__init__` that handles loading them when saved, since it also handles all other operations regarding them.
</blockquote></details>

Yeah I have added that in the new branch so we don't have to manually set it for every dashitem.
___
*[`sinks/dashboard/items/dashboard_item.py` line 56 at r6](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NSHTfw86Nwe_4nrRv3H:-NSHWh9RDXCWlL0B1bIn:b-z7bptv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/016ab6078e5e59d70b593a70851209c92ecf7c4e/sinks/dashboard/items/dashboard_item.py#L56)):*

This logic should live in the dashboard save/load functions which just call to `get_parameters`. Currently this gives two redundant ways to get the parameters which doesn't make a ton of sense imo. Also remember that `parameter.saveState()` and `parameter.restoreState()` exist and probably work better for the save/load logic.
</blockquote></details>

I was going to do that but it does not work with the "series" parameter. Basically `saveState` stores everything, including the options, i.e. the list of streams, and it loads them in the same way. The problem here is that the list of streams are subject to change, which leads in undefined behaviour and sometimes crash.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 77 at r5](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-NS8uOtd9-0EeqpFTtzc:-NSHgfhD0VfWTUjjtdU7:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/bc3426ca9aa1802625c7bb70c86ea9814e8c4852/sinks/dashboard/items/plot_dash_item.py#L77)):*

This is the more general solution (dashitems might want to do more than just be recreated on param change), I think we should leave it as is for now and once we have a better selection of dashitems we can decide whether the recreation approach is sufficient.
</blockquote></details>

Done.

KavinSatheeskumar (commented):
1. Yes we are removing `prompt_for_properties` in this PR, since we are already part of the way through that may as well finish
2. Yes, we want pos in the parameters, as its nice to be able to specify exactly where things go (appologies for the pain this will cause)
3. Having a handler "on_props_change" is more extensible, so yeah don't create and delete, just have an on_props_change

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/104#-:-NSNc_5j032G0Ql1Xs1O:b-it02nz)* status: all files reviewed, 11 unresolved discussions (waiting on @BluCodeGH and @itslinotlie)

Issue #134, opened 2023-09-16
Omnibus Version Number Updating


Darth-Raazi:
Since its creation, Omnibus' version number has remained at `0.0.1`. This needs to get updated to the current version (let's say `1.0.0`). Also some logic to auto-increment this number would be nice, maybe every PR increments the minor version number and every merge into `master` increments the major version number

PR #160, opened 2023-10-24
Implement Auto-Increment of Major Version for Omnibus Library

shirleyfyx:
- Introduced a GitHub Actions workflow that automates the process of incrementing the major version number of the Omnibus library. 

- The workflow is designed to trigger pull requests that introduce changes to files within the omnibus/ directory on the main branch.

- Closes https://github.com/waterloo-rocketry/omnibus/issues/134

### DIFF
--- a/.github/workflows/increment-major-version.yml
+++ b/.github/workflows/increment-major-version.yml
@@ -1,0 +1,44 @@
+name: Increment Major Version
+
+on:
+  pull_request:
+    paths:
+      - 'omnibus/**'
+    branches:
+      - main
+
+jobs:
+  increment-version:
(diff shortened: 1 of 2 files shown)


shirleyfyx:
The major version number is able to increment automatically.  
image
image

shirleyfyx:
After careful deliberation, we have resolved to adopt a manual approach for updating version numbers, acknowledging that variations are necessary to accurately reflect the nature of changes made. Consequently, I will be establishing a GitHub Actions workflow dedicated to validating whether the version number has been updated, and if not, it will trigger a reminder notification.

PR #197, opened 2024-01-22
Data post-processing script rewrite

ManavToor:
# For kavin to roast

## Description

Rewrote old script cleaner

Object oriented the code

n/a

## Developer Testing

- run the code, no args 

## Reviewer Testing

- Run main.py and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/197)


tools/main.py
```diff
@@ -0,0 +1,131 @@
...
+                previous_timestamp = packet.timestamp
+
+    # Remove the initial entry
+    del data[0]
+
+    data_list = list()
+    for timestamp, items in data.items():
+        data[timestamp]['Timestamp'] = timestamp
+        data_list.append(items)
+    del data
```
KavinSatheeskumar, line 55: why are we calling `del data`?

tools/main.py
```diff
@@ -0,0 +1,131 @@
...
+
+    return data_list
+
+
+def to_csv(outfile, field_names, data):
```
KavinSatheeskumar, line 60: use pandas for this instead of your own helper function

tools/main.py
```diff
@@ -0,0 +1,131 @@
...
+        counter += 1
+
+    # User selects what boards' data they would like to read, those boards' info
+    # from 'available' is added to 'selected'.
+    # 'user_in' (user input as a list) is deleted to save space
+    user_in = input('Enter comma-seperate what values').split(', ')
+    selected = list()
+    for item in user_in:
+        selected.append(available[int(item)])
+    del user_in
```
KavinSatheeskumar, line 110: we do we del user_in here?

tools/main.py
```diff
@@ -0,0 +1,131 @@
...
+
+def main(infile):
+    # Read msgpack file
+    infile = open(infile, 'rb')
+
+    # Search log file to see which sensors and actuators have readings logged.
+    # Turns each logged packet into a Packet object for simplicity.
+    # Adds all found boards into 'available' set
+    available = set()
+    for raw_packet in msgpack.Unpacker(infile):
```
KavinSatheeskumar, line 89: we read the file twice (once here and once in retrieve data), it would be nice if we could wrap this all into one class and have the constructor read the file and some other methods get us the values we need when we need them

KavinSatheeskumar (changes requested):

tools/Packet.py
```diff
@@ -0,0 +1,82 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Packet:
```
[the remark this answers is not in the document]
ManavToor: Renamed to Message

tools/main.py
ManavToor: not ideal in this case

SuperZooper3:
Closing to use work on the [higher up branch](https://github.com/waterloo-rocketry/omnibus/tree/aaumais/185-data-post-processing-script-rewrite) with a more modular style.

PR #210, opened 2024-02-15
Log data extraction rewrite (for Flight Dynamics applications)

### LINKED ISSUE #185: Data post-processing script rewrite
I had a chat with Jack last term about possible quality of life improvements to the post-processing script (`tools/data_processing`). The goals I've identified are.

1. Abstract out the hard-coded fields into object-oriented classes instead of the huge if-else tree currently there, to more easily define what fields are being listened for
2. Define the different classes for CAN data and DAQ, and standardize the way discrete vs continuous data are handled through classes
3. Make get-range more interactive, showing CAN data along with DAQ data in the timeline when looking to crop things out, as well as letting you, return from the plot shown and ask for different curves before deciding timestamps

Other nice-to-have features would be:

- Add an option to re-zero the timestamps on data entries when being turned into a CSV so that the 0 is the start of the window selected
- Add some kind of nicer way to browse the data path to find log files to use (a command line file browser maybe?)

This would also come with adding a documentation README to make the script accessible to everyone.

SuperZooper3:
## Description
Implemented functionally similar to the old data extractor, letting us read fields from the logs output from global logger. 

- Made extractor script
- Modularized preview and export functionality for easy future additions 
- Make field peeking script to look at the CAN fields that appear in a log, to select which to use

This PR closes #185 .

## Developer Testing

- Ran the extractor on fake logs generated with fakeni / fakeparsley as well as on real logs from SAC/
- Extracted various different fields, combinations of fields, and for fields that are empty sometimes

## Reviewer Testing

- Read the README.md in the tool's folder
- Download SAC's flight.log from sharepoint https://uofwaterloo.sharepoint.com/:f:/r/sites/tm-eng-waterloorocketry/Shared%20Documents/Comp%20Logs/2023%20SAC?csf=1&web=1&e=ulQauc
- Run the main.py in preview mode with any columns you want, and note down start and end times
- Run the main again in extract mode, pick some DAQ and some CAN columns and see the CSVs and manifest be exported!

**I'm looking for feedback on usability mainly and code quality**

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)


tools/data_processing/main.py
```diff
@@ -1,127 +1,278 @@
...
+
+    if mode == "a" or mode == "c":
+        with open(can_export_path, "w") as outfile:
+            writer = csv.writer(outfile)
+            writer.writerow(["time"] + can_cols)
+            for line in can_data:
+                writer.writerow(line)
+        can_size = os.path.getsize(can_export_path)
+        formatted_can_size = "{:.2f} MB".format(can_size / (1024 * 1024))
+        print(f"CAN data exported to {can_export_path} with size {formatted_can_size}")
```
ManavToor, line 222: perhaps this can be its own function

tools/data_processing/main.py
```diff
@@ -1,127 +1,278 @@
...
+        if int(can_data[i][0]) > int(can_data[i+1][0]):
+            print(f"Warning: CAN timestamp {can_data[i][0]} is greater than {can_data[i+1][0]}")
+
+    return selected_daq_cols, selected_can_cols, daq_data, can_data
+
+
+def data_preview(file_path, mode="a",msg_packed_filtering="behind_stream"):
+    print(f"Previewing {file_path} in mode {mode}")
+    # Modes: a for all, d for daq, c for can
+    if mode != "a" and mode != "d" and mode != "c":
```
ManavToor, line 127: can be rewritten as:
`if mode not in {"a", "d", "c"}:`

tools/data_processing/main.py
```diff
@@ -1,127 +1,278 @@
...
+    msg_packed_filtering_mode = "ahead_stream"
+    if args.behind:
+        msg_packed_filtering_mode = "behind_stream"
+
+    if processing_mode == "p":
+        data_preview(in_file_path, data_mode,msg_packed_filtering=msg_packed_filtering_mode)
+    elif processing_mode == "e":
+        data_export(in_file_path, data_mode,msg_packed_filtering=msg_packed_filtering_mode)
+    else:
+        raise NotImplementedError
```
ManavToor, line 278: All of this would be neater in its own function or class imo

KavinSatheeskumar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: 0 of 10 files reviewed, 13 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, and @SuperZooper3)

___
*[`tools/data_processing/can_field_definitions.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHuVL32-UxTeqJm_Z:-NrlHuVL32-UxTeqJm__:bul0lev) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L13)):*
> ```Python
> 
> 
> class CanProcessingField:
> ```

Add some comments to this class so I can know what all the API is for

Also, type hints would be appreciated.

Note, I am wondering if this is similar to the stuff that goes on in parsley IE should this be made apart of the "field" class for parsley, or maybe the message class
___
*[`tools/data_processing/can_field_definitions.py` line 58 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHbea6nWHzjduyVHj:-NrlHbea6nWHzjduyVHk:b-oq2io3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L58)):*
> ```Python
> # Then you define the CSV name, paste the signature from the discovery script, and choose a value from the dictionary you want to export
> 
> CAN_FIELDS = [
> ```

If possible, use parsley as the source of truth for CAN fields, All this data should be stored in parsley.
___
*[`tools/data_processing/can_field_definitions.py` line 103 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlI8FkAV6pLUvi-RuQ:-NrlI8FkAV6pLUvi-RuR:biwdbmy) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L103)):*
> ```Python
>     TESTING = parser.parse_args().test
> 
>     if not TESTING:
> ```

use pytest for dis :pray: 
___
*[`tools/data_processing/can_field_definitions.py` line 176 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHWkJ9Rx-KwozOiAu:-NrlHWkJ9Rx-KwozOiAv:b-d3xvkv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L176)):*
> ```Python
>     assert not cmatch_cread.match(missing_data_candidate)
> 
>     print("All tests passed!")
> ```

Use pytest for tests, not just if __name__ = "__main__". Pytest is already set up for the omnibus repo
___
*[`tools/data_processing/can_processing.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlIkSWEad6eoth_ZRc:-NrlIkSWEad6eoth_ZRd:b-v05bt5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_processing.py#L10)):*
> ```Python
> 
> 
> def get_can_cols(infile) -> List[str]:
> ```

You can look at the set of columns defined in parsley (IE the fields of all the messages defined in MESSAGES object). This saves you a pass of the data, which probs buys some speed and RAM.
___
*[`tools/data_processing/can_processing.py` line 64 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlID0h6gOw2qbAs6ah:-NrlID0h6gOw2qbAs6ai:bo0s8c7) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_processing.py#L64)):*
> ```Python
>     print("This file is not meant to be run directly. Run main.py instead.")
> 
>     # testing code
> ```

Use pytest, don't use commented out code
___
*[`tools/data_processing/main.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJTis-YPdehZb60Xd:-NrlJTitEsIaXYzCOMpR:b-dt525i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L40)):*
> ```Python
> 
> 
> def ingest_data(file_path, mode="a", daq_compression=True, daq_aggregate_function="average",msg_packed_filtering="behind_stream"):
> ```

Consider using Pandas to do alot of this. I noticed alot of simple row/col ops such as replacing nulls, etc. Pandas was built for that kind of thing. Pandas might also have support for msg pack, which could save you alot of work decompressing.
___
*[`tools/data_processing/main.py` line 71 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJuZbDpOqgx36eUMZ:-NrlJuZbDpOqgx36eUM_:b-kuqmbu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L71)):*
> ```Python
>         "Enter the numbers for the columns you want to extract, seperated by commas, or leave empty for all: ")
> 
>     # parse the selection into a list of indexes in the cols list
> ```

This is me just abusing python syntax, but I personally prefer the below



_Code snippet:_
```Python
indexes = [int(i) for i in selection.replace(" ", "").split(",")] if selection else list(range(1, len(column_mapping) +1))
```
___
*[`tools/data_processing/main.py` line 93 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlKO4s62qf7uB4ua-f:-NrlKO4s62qf7uB4ua-g:b-93rayf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L93)):*
> ```Python
> 
>     # get the data for the selected columns
>     with open(file_path, "rb") as infile:
> ```

pandas can read msg pack, which might save you alot of code

https://pandas.pydata.org/pandas-docs/version/0.25/reference/api/pandas.read_msgpack.html
___
*[`tools/data_processing/old_main.py` line 1 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlKcF6CKd3DAd2OsEb:-NrlKcF6CKd3DAd2OsEc:b-uk7x00) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/old_main.py#L1)):*
> ```Python
> # Process log files (either from the NI source or globallog sink) into .csv
> ```

Just obliterate this, we have git for a reason

SuperZooper3 (commented):
> Dismissed @KavinSatheeskumar from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: 0 of 10 files reviewed, 12 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, @KavinSatheeskumar, @ManavToor, and @SuperZooper3)

___
*[`tools/data_processing/can_field_definitions.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHuVL32-UxTeqJm_Z:-Ns5TDVI77leTOjngP2K:b-f6vjd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L13)):*

Add some comments to this class so I can know what all the API is for

Also, type hints would be appreciated.

Note, I am wondering if this is similar to the stuff that goes on in parsley IE should this be made apart of the "field" class for parsley, or maybe the message class
</blockquote></details>

Working on it for the type hints and comments :thumbsup: 

See below comment for why this is distinct from the parsley stuff, but down to have an architecture chat if you want to.
___
*[`tools/data_processing/can_field_definitions.py` line 58 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHbea6nWHzjduyVHj:-Ns5SK7xAQrfLRBuIdlt:b-4x83qb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L58)):*

If possible, use parsley as the source of truth for CAN fields, All this data should be stored in parsley.
</blockquote></details>

The way i'm currently thinking about how these field definitions, is that they are like abstractions on top of the core message types and definitions. Like we know the pneumatics pressure is the signal coming from the SENSOR_PRESSURE_PNEUMATICS with message type SENSOR_ANALOG, and we want the value from that message. It's different from being a 1 to 1 copy and paste of all the possible messages sent over CAN, given that the current scope for this work is flight dynamics data collection where we know what we need to extract.

To make it less annoying to get the fields, I also included the field_peeking.py script like mentioned in the comment above this code, but like lmk if this makes sense and what you think about it.
___
*[`tools/data_processing/can_field_definitions.py` line 103 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlI8FkAV6pLUvi-RuQ:-Ns5Q_by5-xYtmy_MzA2:b-x8er5u) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L103)):*

use pytest for dis :pray:
</blockquote></details>

Transferring this to a pytest suite has been opened as a new issue to be dealt with after merging this. https://github.com/waterloo-rocketry/omnibus/issues/211 
___
*[`tools/data_processing/can_field_definitions.py` line 176 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHWkJ9Rx-KwozOiAu:-Ns5RGqs0SqxkIptCbZK:b-1ldtxb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L176)):*

Use pytest for tests, not just if __name__ = "__main__". Pytest is already set up for the omnibus repo
</blockquote></details>

See above comment on issue #211
___
*[`tools/data_processing/can_processing.py` line 10 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlIkSWEad6eoth_ZRc:-Ns5RThE-R9b6trbN5-j:bfj66oj) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_processing.py#L10)):*

You can look at the set of columns defined in parsley (IE the fields of all the messages defined in MESSAGES object). This saves you a pass of the data, which probs buys some speed and RAM.
</blockquote></details>

Speed isn't that much of concern, and plus this code here is trying to figure out which columns defined in parsley actually appear in this dataset.
___
*[`tools/data_processing/can_processing.py` line 64 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlID0h6gOw2qbAs6ah:-Ns5Rn442hNLMdpNCr_y:b-43uush) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_processing.py#L64)):*

Use pytest, don't use commented out code
</blockquote></details>

See #211 
___
*[`tools/data_processing/main.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJTis-YPdehZb60Xd:-Ns5Thtc9-K24k58MYm1:b-xnbkik) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L40)):*

Consider using Pandas to do alot of this. I noticed alot of simple row/col ops such as replacing nulls, etc. Pandas was built for that kind of thing. Pandas might also have support for msg pack, which could save you alot of work decompressing.
</blockquote></details>

Yeah, I agree this is sorta poorly done right now. Pandas msgpacked is deprecated though form a quick google search: https://pandas.pydata.org/pandas-docs/version/0.25/reference/api/pandas.read_msgpack.html

Maybe I should just invest more time into abstract out the data cleaning up?
___
*[`tools/data_processing/main.py` line 71 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJuZbDpOqgx36eUMZ:-Ns5TzGz2KLcJzhl72OZ:b-mio1sp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L71)):*

This is me just abusing python syntax, but I personally prefer the below
</blockquote></details>

Yeah that would work but i think the current way is more readable and clear.
___
*[`tools/data_processing/main.py` line 93 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlKO4s62qf7uB4ua-f:-Ns5URVe-uiTkm52QJSX:bwn0zv9) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L93)):*

pandas can read msg pack, which might save you alot of code

https://pandas.pydata.org/pandas-docs/version/0.25/reference/api/pandas.read_msgpack.html
</blockquote></details>

Deprecated since version 0.25.0.
___
*[`tools/data_processing/main.py` line 127 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#gh-1493819046:-Ns5UZRN4cISMZDwZoVo:b-no7cv4) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L127)):*
<details><summary><i>Previously, ManavToor (ManavToor) wrote…</i></summary><blockquote>

can be rewritten as:
`if mode not in {"a", "d", "c"}:`
</blockquote></details>

Could be, but to stay consistent with the rest of the code that uses one or two == / != i think its overall better to leave it as three ifs.
___
*[`tools/data_processing/main.py` line 222 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#gh-1493819023:-Ns5VL_t4FmQn5LxYRN0:b-hcmfbo) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L222)):*
<details><summary><i>Previously, ManavToor (ManavToor) wrote…</i></summary><blockquote>

perhaps this can be its own function
</blockquote></details>

Done, this is lowkey super ugly lol MB.
___
*[`tools/data_processing/main.py` line 278 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#gh-1493819107:-Ns5VROI4ROsPmBggr0Z:bsqebtf) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L278)):*
<details><summary><i>Previously, ManavToor (ManavToor) wrote…</i></summary><blockquote>

All of this would be neater in its own function or class imo
</blockquote></details>

Done, just broke out the arugment passing
___
*[`tools/data_processing/old_main.py` line 1 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlKcF6CKd3DAd2OsEb:-Ns5YT_a9WVxT3-FegOs:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/old_main.py#L1)):*

Just obliterate this, we have git for a reason
</blockquote></details>

Done.

KavinSatheeskumar (dismissed):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: 0 of 10 files reviewed, 19 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, @ManavToor, and @SuperZooper3)

___
*[`tools/data_processing/can_field_definitions.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHuVL32-UxTeqJm_Z:-Ns6S6k6EXuOE-mbsK7B:bgxmfoc) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L13)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Working on it for the type hints and comments :thumbsup: 

See below comment for why this is distinct from the parsley stuff, but down to have an architecture chat if you want to.
</blockquote></details>

I agree with your opinions below.

This might just be me being dumb, but there is only ever one reading signature, so why do we never need matching_pattern to have more than one pair?
___
*[`tools/data_processing/can_field_definitions.py` line 58 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHbea6nWHzjduyVHj:-Ns6QWK5DOGEduA7dsDV:b-o92ieb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L58)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

The way i'm currently thinking about how these field definitions, is that they are like abstractions on top of the core message types and definitions. Like we know the pneumatics pressure is the signal coming from the SENSOR_PRESSURE_PNEUMATICS with message type SENSOR_ANALOG, and we want the value from that message. It's different from being a 1 to 1 copy and paste of all the possible messages sent over CAN, given that the current scope for this work is flight dynamics data collection where we know what we need to extract.

To make it less annoying to get the fields, I also included the field_peeking.py script like mentioned in the comment above this code, but like lmk if this makes sense and what you think about it.
</blockquote></details>

Upon looking at this more closely, I agree with you. There is a lot of info here that isn't CAN message definitions.
___
*[`tools/data_processing/can_processing.py` line 29 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6SlqS-02_ihG9Z73r:-Ns6SlqS-02_ihG9Z73s:bmic428) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/can_processing.py#L29)):*
> ```Python
> 
> 
> def get_can_lines(infile, cols=[],msg_packed_filtering="behind_stream") -> List[List[Union[int, str]]]:
> ```

Sorry to keep hammering on this, but instead of returning a `List[List[Union[int, str]]]`, might it be worth it to return a pandas dataframe?
___
*[`tools/data_processing/main.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJTis-YPdehZb60Xd:-Ns6TJVBDn3bXbYR5aYD:bi9tm5g) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L40)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Yeah, I agree this is sorta poorly done right now. Pandas msgpacked is deprecated though form a quick google search: https://pandas.pydata.org/pandas-docs/version/0.25/reference/api/pandas.read_msgpack.html

Maybe I should just invest more time into abstract out the data cleaning up?
</blockquote></details>

I think we don't need to worry about using pandas to read from msgpack, but using it to handle all the row/col ops + generall working with the data is the right move. 
___
*[`tools/data_processing/main.py` line 127 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#gh-1493819046:-Ns6V8D6Bw_nuXiJD5YA:b-y86sk) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L127)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Could be, but to stay consistent with the rest of the code that uses one or two == / != i think its overall better to leave it as three ifs.
</blockquote></details>

nah, I don't feel like the consistency is worth it here. Not a huge deal tho, so I'm not worried.
___
*[`tools/data_processing/main.py` line 2 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6XPDq6iD8Oax1UE-7:-Ns6XPDq6iD8Oax1UE-8:byjhdb3) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L2)):*
> ```Python
> # A coordinator for the can and daq processors, automatically taking in arguments and running opperations on parsed logs
> # For a single export, it's run a few times for differnt opperations, like preview (with optional ranges) and choosing times, and for exporting the differnt streams with the option to merge
> ```

There are alot of functions in this one file. Maybe try breaking them out into a few files?
___
*[`tools/data_processing/main.py` line 173 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6TiVn7XCVPoCtbs93:-Ns6TiVn7XCVPoCtbs94:b-hlewbi) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L173)):*
> ```Python
> 
> def save_data_to_csv(file_path, data, cols):
>     formatted_can_size = "N/A"
> ```

sorry to keep harping on this, but `pandas.Dataframe.to_csv()` 
___
*[`tools/data_processing/main.py` line 190 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6U8EbBS9FhgWXyg_g:-Ns6U8EbBS9FhgWXyg_h:b-vhnrw5) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L190)):*
> ```Python
> 
>     # The string literal must be un-indented to save properly
>     manifest_text = f"""Data exported from {manifest_args.get("file_path", manifest_empty_filler)} with mode {manifest_args.get("mode", manifest_empty_filler)} and time range {manifest_args.get("start", manifest_empty_filler)} to {manifest_args.get("stop", manifest_empty_filler)}, exported at {datetime.datetime.now()} with the export hash {manifest_args.get("export_hash", manifest_empty_filler)}.
> ```

Save this template in a file, then just load from the file and use .format to format the string before saving to the output file
___
*[`tools/data_processing/main.py` line 203 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6UbTiCgScXnH1X2-D:-Ns6UbTiCgScXnH1X2-E:blcyvfo) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L203)):*
> ```Python
>     """
> 
>     with open(f"{manifest_args.get('file_path', 'NONE').replace('.log','')}_export_{manifest_args.get('export_hash', 'NONE')}_manifest.txt", "w") as manifest_file:
> ```

bit hard to read this big f-strings while they are inside an "open" call. Maybe make a variable immediately above called "filepath", so (in future), if we ever want to change how we save things things, its a bit more obvious how to do so. Minor nitpick
___
*[`tools/data_processing/main.py` line 277 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6WlG-0KvZTJQvchwN:-Ns6WlG-0KvZTJQvchwO:b-ck7ccr) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L277)):*
> ```Python
> 
>     # Create an export hash to identify the export
>     export_hash = hashlib.md5(
> ```

md5 :vomiting_face: (if you are curious, https://link.springer.com/chapter/10.1007/11426639_2)
___
*[`tools/data_processing/main.py` line 295 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6Tu0O0lbxzBVezLIT:-Ns6Tu0O0lbxzBVezLIU:b-qwmymb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L295)):*
> ```Python
> 
>     # save an export manifest for information on what was exported with which settings
>     save_manifest({
> ```

Instead of using a dictionary, use a data class whose purpose is to just hold all these args. It allows for you to be certain of what you are getting and avoid checks like "file_path" in manifest_args
___
*[`tools/data_processing/msgpack_sorter_unpacker.py` line 15 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6VkiwD0WNMi_LsvLG:-Ns6VkiwD0WNMi_LsvLH:bxkr8kz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/msgpack_sorter_unpacker.py#L15)):*
> ```Python
> 
>     # discard all messages if the current time is lower than running current time, by not adding them to the filtered and output list
>     if mode == "ahead_stream":
> ```

You could have the if-statement live inside the for loop, rather than having two copies of the for loop, re=write below



_Code snippet:_
```Python
filtered_messages = []
cur_max_time = 0
for i in range(len(all_messages) - 1):
    if mode == "ahead_stream":
        ...
    elif mode == "behind_stream":
        ...
```

SuperZooper3 (commented):
> Dismissed @KavinSatheeskumar and @ManavToor from 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on @BluCodeGH, @Darth-Raazi, @KavinSatheeskumar, @ManavToor, and @SuperZooper3)

___
*[`tools/data_processing/can_field_definitions.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHuVL32-UxTeqJm_Z:-NsBIqGj6WXNcw0Ovtpw:bhr5dpe) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L13)):*

I agree with your opinions below.

This might just be me being dumb, but there is only ever one reading signature, so why do we never need matching_pattern to have more than one pair?
</blockquote></details>

I'm not sure to have understood your question. We have the matching pattern to check if the message that we're looking at is the one we want, and then we use the reading pattern to extract the data. The reading pattern (which i'll type hint now) is just a single value to read out (if you want multiple values, just make multiple objects). The matching pattern might have multiple pairs though  like board_id': 'SENSOR_VENT', 'msg_type': 'SENSOR_ANALOG', 'data.sensor_id': 'SENSOR_PRESSURE_OX' to uniquely identify one value.
___
*[`tools/data_processing/can_field_definitions.py` line 58 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHbea6nWHzjduyVHj:-NsBIe8N0iLa3CGRWAx_:brkbbr2) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L58)):*

Upon looking at this more closely, I agree with you. There is a lot of info here that isn't CAN message definitions.
</blockquote></details>

Done. (closing this on reviewable)
___
*[`tools/data_processing/can_processing.py` line 29 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6SlqS-02_ihG9Z73r:-NsBPgnC7DVCHkJ9_vj-:bqq8q54) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/can_processing.py#L29)):*

Sorry to keep hammering on this, but instead of returning a `List[List[Union[int, str]]]`, might it be worth it to return a pandas dataframe?
</blockquote></details>

Working on it
___
*[`tools/data_processing/main.py` line 40 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlJTis-YPdehZb60Xd:-NsBQ0fj3ZZKdivjT5_2:bh9d6cs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L40)):*

I think we don't need to worry about using pandas to read from msgpack, but using it to handle all the row/col ops + generall working with the data is the right move.
</blockquote></details>

Working
___
*[`tools/data_processing/main.py` line 93 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlKO4s62qf7uB4ua-f:-NsBTJT-BEVjUIgWQXZ6:b3463pn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L93)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Deprecated since version 0.25.0.
</blockquote></details>

Dismissing for the msgpacked part.
___
*[`tools/data_processing/main.py` line 127 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#gh-1493819046:-NsBUrQ821JVXDEDdcx_:bk57vgd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/main.py#L127)):*

nah, I don't feel like the consistency is worth it here. Not a huge deal tho, so I'm not worried.
</blockquote></details>

Dismissing
___
*[`tools/data_processing/main.py` line 2 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6XPDq6iD8Oax1UE-7:-NsBQ2O98Db6ntwZ9Swz:bh9d6cs) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L2)):*

There are alot of functions in this one file. Maybe try breaking them out into a few files?
</blockquote></details>

Working
___
*[`tools/data_processing/main.py` line 173 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6TiVn7XCVPoCtbs93:-NsBTRogEdN5O0xfcn9B:b-8qu3sp) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L173)):*

sorry to keep harping on this, but `pandas.Dataframe.to_csv()`
</blockquote></details>

that's really saving us like 3 lines, and this function here also lets us calculate the size of the output file which imo is nice to have, but yes if we're going to change it anways i'll just change the saving part to a save_df
___
*[`tools/data_processing/main.py` line 190 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6U8EbBS9FhgWXyg_g:-NsBUBWy6MajPCtdx6bW:b-snvp3e) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L190)):*

Save this template in a file, then just load from the file and use .format to format the string before saving to the output file
</blockquote></details>

Or even better, `.eval` :eyes:  I actually think this is a better solution so that you can have the .get thingys right in the code, rather than worrying about the order which would be annoying im
___
*[`tools/data_processing/main.py` line 203 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6UbTiCgScXnH1X2-D:-NsBTl1a96Ik9yiNfj7E:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L203)):*

bit hard to read this big f-strings while they are inside an "open" call. Maybe make a variable immediately above called "filepath", so (in future), if we ever want to change how we save things things, its a bit more obvious how to do so. Minor nitpick
</blockquote></details>

Done
___
*[`tools/data_processing/main.py` line 277 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6WlG-0KvZTJQvchwN:-NsBV5Bf6bQf1j6PZ4Iv:b-e584ul) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L277)):*

md5 :vomiting_face: (if you are curious, https://link.springer.com/chapter/10.1007/11426639_2)
</blockquote></details>

Hahahah, we dont need a cryptographic hash here. Dismissing
___
*[`tools/data_processing/main.py` line 295 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6Tu0O0lbxzBVezLIT:-NsBW0KuCB772kvNmZRs:bryjzzq) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L295)):*

Instead of using a dictionary, use a data class whose purpose is to just hold all these args. It allows for you to be certain of what you are getting and avoid checks like "file_path" in manifest_args
</blockquote></details>

Working  on it, i'll break it out into its own file
___
*[`tools/data_processing/msgpack_sorter_unpacker.py` line 15 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6VkiwD0WNMi_LsvLG:-NsBXkTID2BHug7cbwO6:bb6un1j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/msgpack_sorter_unpacker.py#L15)):*

You could have the if-statement live inside the for loop, rather than having two copies of the for loop, re=write below
</blockquote></details>

Done

Darth-Raazi (changes requested):
> Reviewed 4 of 10 files at r1, 1 of 2 files at r2, 4 of 9 files at r3, 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: all files reviewed, 16 unresolved discussions (waiting on @BluCodeGH, @KavinSatheeskumar, @ManavToor, and @SuperZooper3)

___
*[`sources/fakeni/main.py` line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrX3cX_Et8RDgYOuOZS:-NrX3cX_Et8RDgYOuOZT:b-psj1i2) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/sources/fakeni/main.py#L12)):*
> ```Python
> 
> READ_BULK = 200  # mimic how the real NI box samples in bulk for better performance
> SAMPLE_RATE = 1000  # total samples/second
> ```

Why was the sample rate decreased?
___
*[`tools/data_processing/can_field_definitions.py` line 108 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCYcGk8pZi7EOxZnyY:-NtCYcGk8pZi7EOxZnyZ:bkgiqrn) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_field_definitions.py#L108)):*
> ```Python
> if __name__ == "__main__":
>     parser = argparse.ArgumentParser(description="Run tests for field_definitions.py")
>     parser.add_argument("-test", action="store_true", help="Run tests")
> ```

Multi-character flags usually use two dashes, e.g. `--test`
___
*[`tools/data_processing/can_field_definitions.py` line 156 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCYk3Q9FZ9IDWdTtj9:-NtCYk3Q9FZ9IDWdTtjA:b-szd31g) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_field_definitions.py#L156)):*
> ```Python
>     print("Testing matching")
>     print("Correct input matching")
>     assert cmatch_cread.match(correct_candidate)
> ```

It's better to have the asserts right after their corresponding test call. That way if the first one fails, the script will exit immediately and not waste time running all the other tests.
___
*[`tools/data_processing/can_processing.py` line 66 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZ-iS8wcEX64P6CRb:-NtCZ-iS8wcEX64P6CRc:b2nhkcb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_processing.py#L66)):*
> ```Python
> 
>     # testing code
>     # uncomment this code to test the file
> ```

Having tests as comments is not the best design. Either have the user pass in a `--test` flag to run tests, or have them defined in a separate file
___
*[`tools/data_processing/daq_processing.py` line 77 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZMnc8cy-AtsA8jLk:-NtCZMnc8cy-AtsA8jLl:b-34vmyd) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/daq_processing.py#L77)):*
> ```Python
> 
>     # testing code
>     # uncomment the following to test this file
> ```

See above
___
*[`tools/data_processing/main.py` line 6 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZykQ4YBbu0qvbLP3:-NtCZykQ4YBbu0qvbLP4:b-1zvwwk) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/main.py#L6)):*
> ```Python
> import argparse
> 
> from interractions import data_preview, data_export
> ```

Nit: `interactions`is misspelled, here and as the filename

SuperZooper3 (commented):
> Dismissed @SuperZooper3, and @SuperZooper3 from 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210)* status: 6 of 17 files reviewed, all discussions resolved (waiting on @BluCodeGH and @Darth-Raazi)

___
*[`tools/data_processing/can_field_definitions.py` line 13 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrlHuVL32-UxTeqJm_Z:-NtpjXpR18KqcrZupqla:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/tools/data_processing/can_field_definitions.py#L13)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

I'm not sure to have understood your question. We have the matching pattern to check if the message that we're looking at is the one we want, and then we use the reading pattern to extract the data. The reading pattern (which i'll type hint now) is just a single value to read out (if you want multiple values, just make multiple objects). The matching pattern might have multiple pairs though  like board_id': 'SENSOR_VENT', 'msg_type': 'SENSOR_ANALOG', 'data.sensor_id': 'SENSOR_PRESSURE_OX' to uniquely identify one value.
</blockquote></details>

Done.
___
*[`tools/data_processing/can_field_definitions.py` line 108 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCYcGk8pZi7EOxZnyY:-NtpjiuA1r1G_9E-F9U_:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_field_definitions.py#L108)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

Multi-character flags usually use two dashes, e.g. `--test`
</blockquote></details>

Done.
___
*[`tools/data_processing/can_field_definitions.py` line 156 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCYk3Q9FZ9IDWdTtj9:-Ntpjn5aCBRtfn29B_-9:b3omnep) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_field_definitions.py#L156)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

It's better to have the asserts right after their corresponding test call. That way if the first one fails, the script will exit immediately and not waste time running all the other tests.
</blockquote></details>

The asserts here are for the actual matching calls, so we're only asserting that these exact calls return what they should. So these are where they should be I think.
___
*[`tools/data_processing/can_processing.py` line 29 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6SlqS-02_ihG9Z73r:-Ntpk0nG0Pe6KPDAV1fj:b-i7gtyh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/can_processing.py#L29)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Working on it
</blockquote></details>

Broken out into sub-PR.
___
*[`tools/data_processing/can_processing.py` line 66 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZ-iS8wcEX64P6CRb:-Ntpk8GA0OacwJvSrhRZ:b-gl2dmt) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_processing.py#L66)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

Having tests as comments is not the best design. Either have the user pass in a `--test` flag to run tests, or have them defined in a separate file
</blockquote></details>

Making a pytest suite is a differnt GH issue that has been broken out, so this should be fine temporarily and serves as a reminder to deal with that issues.
___
*[`tools/data_processing/daq_processing.py` line 77 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZMnc8cy-AtsA8jLk:-NtpkQXh-sdxp6-A0OV1:bsgzb4r) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/daq_processing.py#L77)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

See above
</blockquote></details>

See above answer.
___
*[`tools/data_processing/main.py` line 2 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6XPDq6iD8Oax1UE-7:-NtpkVzW9r-Ae1p4uS3r:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L2)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Working
</blockquote></details>

Done.
___
*[`tools/data_processing/main.py` line 190 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-Ns6U8EbBS9FhgWXyg_g:-NtpkyrQ7HZjn70vPveS:bjxmbro) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/06968c35d2ab053a309ebd1c1d29082875091510/tools/data_processing/main.py#L190)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

Or even better, `.eval` :eyes:  I actually think this is a better solution so that you can have the .get thingys right in the code, rather than worrying about the order which would be annoying im
</blockquote></details>

(this has just been broken out into a differnt F string in it's own python file that works fine)
___
*[`tools/data_processing/main.py` line 6 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCZykQ4YBbu0qvbLP3:-NtpkXtL7sk6iDDIaMe8:bq0ya5i) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/main.py#L6)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

Nit: `interactions`is misspelled, here and as the filename
</blockquote></details>

Whooops.... Done
___
*[`sources/fakeni/main.py` line 12 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NrX3cX_Et8RDgYOuOZS:-Ntpj8z07mAkkDz_9ZpX:bov1xfu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/f331715f67682d9de7325923e6745e339116136e/sources/fakeni/main.py#L12)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

Why was the sample rate decreased?
</blockquote></details>

not sure, I might have been producing fake data. Has been reverted.

Darth-Raazi (approved):
:lgtm:

> Reviewed 11 of 11 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-:-NwWaKY2BvzYXoDZ7cxu:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BluCodeGH and @KavinSatheeskumar)

___
*[`tools/data_processing/can_field_definitions.py` line 156 at r4](https://reviewable.io/reviews/waterloo-rocketry/omnibus/210#-NtCYk3Q9FZ9IDWdTtj9:-NwW_CRJ0nLOcQ_Le8wW:bpc8vzg) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7410583246bcab67a2f8a176fa8a3b93c0c65b40/tools/data_processing/can_field_definitions.py#L156)):*
<details><summary><i>Previously, SuperZooper3 (SuperZooper3) wrote…</i></summary><blockquote>

The asserts here are for the actual matching calls, so we're only asserting that these exact calls return what they should. So these are where they should be I think.
</blockquote></details>

Yeah I was wrong lol, sorry about that. Looks good

PR #234, opened 2024-06-14
Launcher / Log Update 

### LINKED ISSUE #182: Add Argument Support to Launcher
Some sinks and sources need CLI arguments (flags) to be passed in when starting them (ex. `parsley`). Currently, Launcher just calls `sink-or-source/name/main.py`, but the user should also have the option to enter some arguments and choose which source/sink they go to.

patrick-gu:
## Description

This PR closes #182 .

## Developer Testing

- Ran A
- Ran B
- Ran C

## Reviewer Testing

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/234)


zangjiucheng:
I will continue to work for this PR, which I believe is something we really need right now.

zangjiucheng:
## Description

- This is an enhance implement of #182,  where the launcher has been used completely outdate of our code.
- This feature implements the functional use cases of the new version of the launcher.

- Finish implement of launcher arg support based on Patrick previous modify. 
- Support stdout to the screen but not redirect to the log files 
- Support --text mode argument input
- Modify log file name with temp_stamp for permeant storage log for analysis. 
- Tiny fix for log tool with args index for general case



This PR closes #182 .

## Known Issue
- Stdout feature not support `--text` version, (I think to answer each take too much time and useless, you can just do cli then) 
- `--last` command  not support to record stdout flag yet. (Maybe later PR modify this) 

## Developer Testing


- Run any combination with launcher to test if program can functional work.


## Reviewer Testing


- Same as Developer Testing


- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/234)

launcher.py
```diff
@@ -137,14 +147,23 @@ def validate_inputs(self, choices, module):
 
     # Execute commands as subprocesses
     def subprocess(self):
-        print("Launching... ", end="")
+        print(f"Launching... Run commands: {self.commands}")
         for command in self.commands:
+            launch_command = command[:-1]
+            std_flag = command[-1]
```
patrick-gu, line 153: I don't think adding a boolean onto the command is a good way to do this. Perhaps making a class for this, having both the command and the stdout flag, would be cleaner?

sources/replay_log/main.py
```diff
@@ -61,7 +61,7 @@ def get_replay_log(max_logs: int) -> Path | None:
     args = parse_arguments()
     max_logs = args.max_logs
     replay_speed = args.replay_speed
-    log_file = args.log_file if args.log_file != None else get_replay_log(max_logs)
+    log_file = os.path.expanduser(args.log_file if args.log_file != None else get_replay_log(max_logs))
```
patrick-gu, line 64: is there a reason for this?

logtool.py
```diff
@@ -19,11 +20,13 @@ def add_logger(self, filepath=None, core=False):
             file_name, file_dir = file_ref[-1], file_ref[-2]
         else:
             file_name, file_dir = "core_library", ""
-        os.makedirs(os.path.dirname(f'logs/{file_dir}/{file_name}.log'), exist_ok=True)
+        current_time = datetime.now().strftime("%Y%m%d_%H%M")
+        log_file_name = f"{file_name}_{current_time}.log"
+        os.makedirs(os.path.dirname(f'logs/{file_dir}/{log_file_name}'), exist_ok=True)
```
patrick-gu, line 25: looks like a good improvement

launcher.py
```diff
@@ -222,58 +246,85 @@ def __init__(self):
...
         source_list.setLayout(self.src_layout)
         source_list.setContentsMargins(0, 0, 0, 0)
 
+
         #Connect checkbox state to signals to detect which sources were selected 
-        for checkbox in self.src_checkboxes:
-            checkbox.stateChanged.connect(self.update_selected)
+        for i, (checkbox, args) in enumerate(self.src_widgets):
+            # https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result
+            def stateChanged(state, i=i):
```
patrick-gu, line 278: This is really small nitpick. Although the Python documentation officially suggests you can do this, I would be inclined to avoid this. The reason is because we don't know if `.connect` will always only give us one argument, and I assume that nobody really wants to take the time to check (and just looking at the [outdated] docs quickly is unclear https://doc.qt.io/qtforpython-5/PySide2/QtCore/Signal.html#PySide2.QtCore.Signal.connect). I think it's better to make sure it's clear that this is a function with arity 1 (i.e. takes exactly one argument).

launcher.py
```diff
@@ -222,58 +246,85 @@ def __init__(self):
...
+            group_widget = QWidget()
+            group_layout = QVBoxLayout()
+            group_layout.addWidget(checkbox)
+            group_layout.addWidget(std_out)
+            group_widget.setLayout(group_layout)
+            group_widget.setStyleSheet("border: 1px solid black; padding: 5px;")
+            self.sink_layout.addWidget(group_widget, row, col)
             col += 1
-            if col == 3:
+            if col == 5:  # Three groups per row
```
patrick-gu, line 311: I don't really like how this is hardcoded but it's probably fine (too much work to do it cleanly)

patrick-gu (commented):
Left some comments about the code. Also it doesn't seem to work for me, I get the error:

```
Running in GUI mode
Launching... Run commands: [['venv/Scripts/python', '-m', 'omnibus', False], ['venv/Scripts/python', 'sources/fakeni/main.py', False], ['venv/Scripts/python', 'sources/parsley/main.py', '--fake', False], ['venv/Scripts/python', 'sinks/dashboard/main.py', True], ['venv/Scripts/python', 'sinks/printer/main.py', False]]
Traceback (most recent call last):
  File "...\omnibus\launcher.py", line 409, in <module>
    main()
  File "...\omnibus\launcher.py", line 403, in main
    gui_launcher.subprocess()
  File "...\omnibus\launcher.py", line 159, in subprocess
    process = subprocess.Popen(launch_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Program Files\Python312\Lib\subprocess.py", line 1538, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

launcher.py
zangjiucheng: Hi :)) I didn't think about a good way to pass one argument because I have no idea how to edit state binding to a list. But I did use another solution with `partial` function to ensure it take two argument.

zangjiucheng:
I did test this script on the Windows and Linux also, unfortunately I cannot reproduce the issue `[WinError 2] The system cannot find the file specified`. Can you try again please? Maybe it been fixed in some comments I don't really noticed. Thanks for your review :))

patrick-gu:
Sorry I never responded to this! I've figured out that the issue was that I created my venv in a folder called `.venv` instead of `venv`. It shouldn't be an issue since the recommended setup instructions are compatible with this. Also, I haven't tested it thoroughly, but it does seem to work, so thank you :)

zangjiucheng:
@patrick-gu No worries :). I just get some free time today to review some code before. I did test more with more combination of sources and sinks. It should be working well! If any others come up, we can still fix it later lol. Merry Christmas 🎄

PR #255, opened 2024-07-08
Add Unsave Indicator

### LINKED ISSUE #225: Add unsaved file indicator
There should be an indicator somewhere in the app window that tells the user whether they have saved their file or not. This is the same as the little asterisk that appears next to a file name on VScode, indicating to the user that they have not saved yet. When they save, the asterisk/indicator should disappear.

zangjiucheng:
## Description

Changed `dashboard.py` with the following:
- add two properties, `dict[str, Any]:  current_data` and `bool: unsave_indicator`.
- add function `bool: check_for_changes -> return state of change`.
- Call function in `every_second()` since it updates every few milliseconds.
- Tiny formatting of some of the lines.

This PR closes #225.

## Developer Testing

- Test some of the `change`, `save`, and `load` functions to ensure that the unsave indicator is displayed properly.

## Reviewer Testing

- Works fine, let me know if you have any problems :)))

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/255)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -312,7 +312,7 @@ class Dashboard(QWidget):
         # For every 5 second, check if there are any changes
         self.timer = QTimer(self)
         self.timer.timeout.connect(self.change_detector)
-        self.timer.start(1000)  # Check every second
+        self.timer.start(100)  # Check every 0.1 seconds
         
 
     def select_instance(self, name):
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -141,6 +141,7 @@ class Dashboard(QWidget):
 
         # Keep track on whether the save popup should be shown on exit.
         self.should_show_save_popup = True
+        
         # Create a GUI
         self.width = 1100
         self.height = 700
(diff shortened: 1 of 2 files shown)


patrick-gu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/255)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @BigLiban, @Darth-Raazi, @Lucifersan, @saads312, and @zangjiucheng)

___
*[`sinks/dashboard/dashboard.py` line 327 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/255#-O1UtefaDXReW2NvSl_G:-O1UtefaDXReW2NvSl_H:bsqdjqe) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/38aed5dc2a5c5fb31ec8e45991d8ac157af0076b/sinks/dashboard/dashboard.py#L327)):*
> ```Python
>     def every_second(self, payload, stream):
>         # For every second, check if there are any changes
>         if self.check_for_changes():
> ```

I don't think this is the right way to do it because it's only called if there is data attached. An easy fix is to just set a timer to do this. I think the better fix would be to actually mark the file as unsaved (if applicable) every time we make changes but I know that's going to be harder.

zangjiucheng:
Thanks! I'll go with QTimer, updating based on time ensures that the program is more adaptable for future improvements, otherwise we might need to change each type of item to accommodate this little feature. (Since some data is only set in a single dashboard item, it's hard to check each individual change). Also, depending on the time, we don't get errors when switching states :)) that's what I think 💡 Thanks for your comment!

zangjiucheng:
Sorry, I did a wrong push to this branch, but it has been fixed anyway. It's ready to review again lol!

patrick-gu (approved):
I still think the best solution in theory would be to do that, as that would be the most correct/intuitive behaviour. I believe we could do it by monitoring every parameter added, so we wouldn't be able to forget anything.

However I think with the timer it works fine and this is a perfectly reasonable solution. I've set the delay to 100 ms to make it more responsive, and also made it immediately update on saving.

Thanks! I'll merge this now

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/255#-:-O1iiKdYEsO35trKtblJ:bk2oh76)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BigLiban, @Darth-Raazi, @Lucifersan, and @saads312)

zangjiucheng:
I agree with you since this is a tiny feature such that it shouldn't require for much resources to compute, after LC I will open another PR to change this feature follow this idea. But I need some time to implement this and may require each dashboard item to accomplish this feature lol. Thanks for your review!

Darth-Raazi (commented):
Good work on this, one bug I noticed immediately is that if you load from a savefile, it'll display the unsaved indicator which is clearly not right. Not a critical bug though

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/255#-:-O1lvCy5Al6L5h_jfFH3:b-pnkcs4)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

zangjiucheng:
I will fix that associate reopen question on new branch

PR #260, opened 2024-07-11
Theme Switch Button

jeffrey-huang-yz:
…emes, starting from the user's OS theme.

## Description

Added a menu bar button that changes the user's current theme (light or dark)

Added two methods to set the theme using a light or dark palette, and created a variable to detect the user's current theme (based on the OS theme) and apply the correct initial theme. 

This PR closes #XXX.

## Developer Testing

- Ran omnibus and switched the theme and ensuring the theme was applied to all components 
- Ran omnibus beginning with both light and dark themes on my OS to ensure that functionality was consistent 

## Reviewer Testing

Run omnibus and ensure that the themes are sufficient and can be toggled as intended 

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/260)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -6,7 +6,7 @@ import signal
 
 import pyqtgraph
 from pyqtgraph.Qt.QtCore import Qt, QTimer
-from pyqtgraph.Qt.QtGui import QPainter, QAction
+from pyqtgraph.Qt.QtGui import QPainter, QAction, QPalette, QColor
 from pyqtgraph.Qt.QtWidgets import (
     QGraphicsView,
     QGraphicsScene,
@@ -25,7 +25,7 @@ from pyqtgraph.Qt.QtWidgets import (
     QCheckBox,
     QHBoxLayout,
     QPushButton,
-    QGraphicsProxyWidget
+    QGraphicsProxyWidget,
 )
 from pyqtgraph.parametertree import ParameterTree
 from items import registry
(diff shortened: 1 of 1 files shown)


Darth-Raazi (commented):
Good work on implementing this Jeffrey, however I'm not a huge fan of the result. Because of the limited number of things you're manually setting colour for, some visual elements get lost or are hard to see (I especially noticed this with Standard Display Item). I won't block, if @Lucifersan really wants this then you can merge, but I think the better result is achieved by just switching OS theme

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/260#-:-O1UdRCg7ZUMi5zrxnqv:bo4raql)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @BigLiban, @Lucifersan, @patrick-gu, and @saads312)

patrick-gu:
Is there a way we could make this use the styling that would appear if you change OS theme? And then manually override where not available, e.g. I don't think there's a builtin dark theme for Windows

jeffrey-huang-yz:
I could maybe attempt to change the palette so that it's more similar to the result you would see if you were to change the OS but it probably still wouldn't be as perfect... Also, there is a built-in dark theme for Windows already with pyqt (as long as your system is set to dark mode) but it doesn't let you switch to it manually so I believe the only thing you can do is make your own palette

patrick-gu:
This is indeed pretty unfortunate :(

Darth-Raazi:
Closing this PR since the readme was updated. I think we'll stick with OS theme for now, maybe this can be revisited in the future

PR #272, opened 2024-07-15
Update Unsave Indicator for current working file name display

zangjiucheng:
## Description

Changed `dashboard.py`, this is an improvement based on the last [PR 255](https://github.com/waterloo-rocketry/omnibus/pull/255) with unsave indicator.

Features: 
- For the unsave indicator, the top will be associated with the name of the current edit.
- It would make more sense to remove the default loading in `savefile.json`, otherwise it always shows the same name.
- 

This PR closes #255.

## Developer Testing

- Basic `load()`, `save()`, `open()` function calls ensure that names are displayed correctly

## Reviewer Testing

Not a lot different since last PR.

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -10,12 +10,20 @@ import json
 from logtool import Logger
 from pyqtgraph.Qt import QtGui
 from pyqtgraph.Qt.QtWidgets import (
-    QApplication, QDialog, QLabel, QDialogButtonBox, QVBoxLayout,
-    QWidget, QCheckBox, QGridLayout
+    QApplication,
+    QDialog,
+    QLabel,
+    QDialogButtonBox,
+    QVBoxLayout,
+    QWidget,
+    QCheckBox,
+    QGridLayout,
 )
+
 # Some specific commands are needed for Windows vs macOS/Linux
 if sys.platform == "win32":
     from subprocess import CREATE_NEW_PROCESS_GROUP
+
     python_executable = "venv/Scripts/python"
 else:
     python_executable = "python"
@@ -26,10 +34,11 @@ else:
 class Finished(Exception):
     pass
 
+
 # CLI Launcher
 
 
-class Launcher():
+class Launcher:
     def __init__(self) -> None:
         super().__init__()
         self.commands = []
@@ -39,7 +48,7 @@ class Launcher():
         self.load_last: bool = False
 
         # Parse folders for sources and sinks
-        self.modules = {"sources": os.listdir('sources'), "sinks": os.listdir('sinks')}
+        self.modules = {"sources": os.listdir("sources"), "sinks": os.listdir("sinks")}
 
         # Remove dot files
         for module in self.modules.keys():
@@ -54,10 +63,10 @@ class Launcher():
             self.load_last = False
             self.print_choices()
             return
-        with open('lastrun.json', 'r') as fp:
+        with open("lastrun.json", "r") as fp:
             data = json.load(fp)
-            self.src_selected = data['Sources']
-            self.sink_selected = data['Sinks']
+            self.src_selected = data["Sources"]
+            self.sink_selected = data["Sinks"]
             fp.close()
 
     # Print list of sources and sinks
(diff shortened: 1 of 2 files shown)


zangjiucheng:
@Darth-Raazi I am not sure if I fully understand your request, lmk if this is the feature you required

patrick-gu (changes requested):
Works fine, just some things about the implementation

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/272#-:-O28HcDo8p09v8iXGi9s:br3f6sm)* status: all files reviewed, 4 unresolved discussions (waiting on @BigLiban, @Darth-Raazi, @Lucifersan, @saads312, and @zangjiucheng)

___
*[`sinks/dashboard/dashboard.py` line 317 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/272#-O28GsM71ZuvoJN3PaYL:-O28GsM71ZuvoJN3PaYM:b-gslgiz) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7e9f6089596b1f5325588bf48292c87c8ea32c35/sinks/dashboard/dashboard.py#L317)):*
> ```Python
>         self.timer.start(100)  # Check every 0.1 seconds
> 
>         self.file_name = None # Current file name
> ```

I think we should compute this as necessary instead of making a state here, as that would be more reliable
___
*[`sinks/dashboard/dashboard.py` line 333 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/272#-O28HBKo6-FTzGzrB6r2:-O28HBKo6-FTzGzrB6r3:b-y0zxd1) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7e9f6089596b1f5325588bf48292c87c8ea32c35/sinks/dashboard/dashboard.py#L333)):*
> ```Python
>         return False
> 
>     def change_detector(self):
> ```

Maybe we should call this explicitly when we change the filename? Not a big deal though as it's fast enough as it is. Just would be nice to get some immediate feedback
___
*[`sinks/dashboard/dashboard.py` line 339 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/272#-O28HKWB-kSb80lnHUdL:-O28HKWB-kSb80lnHUdM:b-qsts0e) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7e9f6089596b1f5325588bf48292c87c8ea32c35/sinks/dashboard/dashboard.py#L339)):*
> ```Python
>             title_string += f" - {self.file_name}"
>         else:
>             title_string += " - Untitled"
> ```

If you do this based on `self.file_location` then you don't need to make this untitled - it can be just savefile.json, unsaved.
___
*[`sinks/dashboard/dashboard.py` line 910 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/272#-O28HWlL2-dN2rgda50H:-O28HWlL2-dN2rgda50I:b-hyoy0j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/201ecc2b510c4d9149ef15c822ff7931063d76e2/sinks/dashboard/dashboard.py#L910)):*
> ```Python
>     dash.update()
>     dash.show()
>     dash.load()
> ```

Why was this removed?

zangjiucheng:
Thanks for the suggestion, here's an updated version of my code, as we said, I was replacing `file_location` with the `file_name` parameter.

zangjiucheng:
This update is a quick update after the time state change. 

Also, I bring back the `load()` function which is used to default load `savefile.json`. I am not sure if we should default load it or wait to use the load function from the Qt UI. 

Lmk for your thoughts.

patrick-gu (changes requested):
I think there are still some inconsistencies, e.g. if you lock and unlock the dashboard it doesn't show the savefile. I think you should check all usages of `self.setWindowTitle`

zangjiucheng:
Since it's a long term not update this.... I am decide to give up this PR and reimplement this feature again later.

PR #274, opened 2024-07-16
Toggle Mouse Resizing

Darth-Raazi:
## Description

Added toggle to enable/disable mouse resizing, cleaned up dashboard locking, cleaned up menu bar

This PR closes #XXX.

## Developer Testing

- Ran A
- Ran B
- Ran C

## Reviewer Testing

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274)

### DIFF
--- a/sinks/dashboard/dashboard.py
+++ b/sinks/dashboard/dashboard.py
@@ -132,6 +132,9 @@ class Dashboard(QWidget):
         # Keep track of if editing is allowed
         self.locked = False
 
+        # Keep track of whether mouse resizing is allowed
+        self.mouse_resize = True
+
         # Determine the specific directory you want to always open
         script_dir = os.path.dirname(os.path.abspath(__file__))
         self.save_directory = os.path.join(script_dir, "..", "..", "sinks", "dashboard", "saved-files")
@@ -204,27 +207,15 @@ class Dashboard(QWidget):
             new_action.triggered.connect(create_registry_trigger(i))
             self.lockableActions.append(new_action)
 
-        # adding a button to the dashboard that removes all dashitems on the screen
-        remove_dashitems = menubar.addMenu("Clear")
-        remove_dashitems_action = remove_dashitems.addAction("Remove all the dashitems")
-        remove_dashitems_action.triggered.connect(self.remove_all)
-        self.lockableActions.append(remove_dashitems_action)
-
-        # adding a button to switch instances of parsley
-        self.can_selector = menubar.addMenu("Parsley")
-
         # Add an action to the menu bar to lock/unlock
         # the dashboard
-        add_lock_menu = menubar.addMenu("Lock")
-        lock_action = add_lock_menu.addAction("Lock Dashboard (^l)")
-        lock_action.triggered.connect(self.lock)
-        self.lockableActions.append(lock_action)
-        unlock_action = add_lock_menu.addAction("Unlock Dashboard (^l)")
-        unlock_action.triggered.connect(self.unlock)
-        lock_selected = add_lock_menu.addAction("Lock Selected (l)")
+        editing_menu = menubar.addMenu("Editing")
+        self.lock_action = editing_menu.addAction("Lock Dashboard (^l)")
+        self.lock_action.triggered.connect(self.toggle_lock)
+        lock_selected = editing_menu.addAction("Lock Selected (l)")
         lock_selected.triggered.connect(self.lock_selected)
         self.lockableActions.append(lock_selected)
-        self.unlock_items_menu = add_lock_menu.addMenu("Unlock Items")
+        self.unlock_items_menu = editing_menu.addMenu("Unlock Items")
         """Menu containing actions to unlock items that are locked, in the order
         in which they were locked.
         """
@@ -233,13 +224,21 @@ class Dashboard(QWidget):
         were locked.
         
         Includes the rect item and the unlock action."""
+        self.mouse_resize_action = editing_menu.addAction("Mouse Resizing (^m)")
+        self.mouse_resize_action.setCheckable(True)
+        self.mouse_resize_action.setChecked(True)
+        self.mouse_resize_action.triggered.connect(self.toggle_mouse)
+        self.lockableActions.append(self.mouse_resize_action)
 
         # An action to the to the menu bar to duplicate
         # the selected item
-        duplicate_item_menu = menubar.addMenu("Duplicate")
-        duplicate_action = duplicate_item_menu.addAction("Duplicate Item (^d)")
+        items_menu = menubar.addMenu("Items")
+        duplicate_action = items_menu.addAction("Duplicate Item (^d)")
         duplicate_action.triggered.connect(self.on_duplicate)
         self.lockableActions.append(duplicate_action)
+        remove_action = items_menu.addAction("Remove All (^r)")
+        remove_action.triggered.connect(self.remove_all)
+        self.lockableActions.append(remove_action)
 
         # We have a menu in the top to allow users to change the stacking order
         # of the selected items.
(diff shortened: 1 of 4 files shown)


patrick-gu (changes requested):
Some comments about impl.

Also, I think mouse resizing remains very weird:
- If you click on something, it'll follow you around, right? It keeps doing this if you turn off mouse resizing
- Or if you click on something then remove all, your mouse still has the grab hand
- The resize corners are still there

Finally I think remove all items is pretty dangerous to have, especially with a keyboard shortcut. A confirmation dialog would probably be useful.

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-:-O1tcZgJ4wsbHu-a51-V:b2qjsb5)* status: all files reviewed, 3 unresolved discussions (waiting on @BigLiban, @Darth-Raazi, @Lucifersan, and @saads312)

___
*[`sinks/dashboard/dashboard.py` line 342 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tb56c52JNV_MdEWMv:-O1tb56c52JNV_MdEWMw:b-n69dsu) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/dashboard.py#L342)):*
> ```Python
>             self.setWindowTitle(f"{title} {unsaved_symbol}")
>         elif not changed and unsaved_symbol in title:
>             self.setWindowTitle(title[:-2])
> ```

what is the point of this? I feel like it would be more robust to just check if the title is equal to the desired title if you are concerned about the performance of changing the window title
___
*[`sinks/dashboard/dashboard.py` line 594 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tbzXYCoSURO0x_U_M:-O1tbzXYCoSURO0x_U_N:b-w2ivtv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/dashboard.py#L594)):*
> ```Python
>         self.load()
> 
>     def toggle_lock(self):
> ```

I don't think it's simpler to merge lock and unlock into one method. Keeping it separate makes it easier to reason about. If you put them together, then it's difficult to keep track of the boolean expressions used
___
*[`sinks/dashboard/utils.py` line 82 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tbMi22sKbgjXiR3gY:-O1tbMi22sKbgjXiR3gZ:bfzbwqb) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/utils.py#L82)):*
> ```Python
>                     self.send_backward.emit()
>                 case KeyEvent(Qt.Key_R, Qt.ControlModifier):
>                     self.remove_all.emit()
> ```

I think it's dangerous to have a remove all keyboard shortcut that could be easily pressed by accident

Darth-Raazi (commented):
I think I fixed most of the issues you mentioned. What exactly do you mean by "it'll follow you around"? Also added a confirm popup for remove all

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-:-O2hN2EV27vLwt_ybHa4:bk8wyfg)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @BigLiban, @Lucifersan, @patrick-gu, and @saads312)

___
*[`sinks/dashboard/dashboard.py` line 342 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tb56c52JNV_MdEWMv:-O2hNG5H6hgpfQAERU1F:bk8hcy0) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/dashboard.py#L342)):*
<details><summary><i>Previously, patrick-gu (Patrick Gu) wrote…</i></summary><blockquote>

what is the point of this? I feel like it would be more robust to just check if the title is equal to the desired title if you are concerned about the performance of changing the window title
</blockquote></details>

So the title text only has to be hardcoded in one place. Otherwise if you update the title, you need to change multiple lines of code
___
*[`sinks/dashboard/dashboard.py` line 594 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tbzXYCoSURO0x_U_M:-O2hNRZGASvdEjmSBq5O:b-94di2j) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/dashboard.py#L594)):*
<details><summary><i>Previously, patrick-gu (Patrick Gu) wrote…</i></summary><blockquote>

I don't think it's simpler to merge lock and unlock into one method. Keeping it separate makes it easier to reason about. If you put them together, then it's difficult to keep track of the boolean expressions used
</blockquote></details>

I disagree. Most of the code is the exact same so it makes sense to combine then
___
*[`sinks/dashboard/utils.py` line 82 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tbMi22sKbgjXiR3gY:-O2hN_92-L_EZnJ85vCG:b-896fix) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/utils.py#L82)):*
<details><summary><i>Previously, patrick-gu (Patrick Gu) wrote…</i></summary><blockquote>

I think it's dangerous to have a remove all keyboard shortcut that could be easily pressed by accident
</blockquote></details>

Done.
___
*[`sinks/dashboard/items/plot_dash_item.py` line 180 at r2](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O2hNdG_CABvWxHLbLMz:-O2hNdG_CABvWxHLbLN-:b-ha2kbh) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/ec2d7727ca1f133f230643c3425d5f439be04045/sinks/dashboard/items/plot_dash_item.py#L180)):*
> ```Python
> 
>     def dynamic_corner_size(self)-> int | float :
>         return 0
> ```

Removing this since mouse resizing can be disabled globally now

patrick-gu (approved):
Sorry, I meant if you short click on it, when you move your mouse around it will follow your mouse even when not held down. Thanks for the confirm popup. I think this looks good, except I will turn off mouse resizing by default due to the high volume of complaints.

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-:-O2mAh-7CteirM4d4sY6:b-e1z1kb)* status: all files reviewed, 2 unresolved discussions (waiting on @BigLiban, @Lucifersan, and @saads312)

___
*[`sinks/dashboard/dashboard.py` line 342 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/274#-O1tb56c52JNV_MdEWMv:-O2m9agy-P6jGcPSQjW2:bqmdji6) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/b0f683cda4c6cbbb8df3db882ea4d8d9f6c89ab5/sinks/dashboard/dashboard.py#L342)):*
<details><summary><i>Previously, Darth-Raazi (Ozayr Raazi) wrote…</i></summary><blockquote>

So the title text only has to be hardcoded in one place. Otherwise if you update the title, you need to change multiple lines of code
</blockquote></details>

IMO, the data flow should be one way, from our fields in this class to the window itself. But I won't block on this.

PR #278, opened 2024-07-26
Refresh limits of series parameters

patrick-gu:
## Description

When you click on a widget, it opens the parameter tree. Now, when that is opened, any parameters which reference series will refresh.

This is definitely not the cleanest way to implement this but I can't think of a better quick solution right now.

## Developer Testing

- Click on a widget, then run parsley, then click away and click again

## Reviewer Testing

- This is more for me but would be good to test this on a replay log I think so it doesn't crash

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/278)

### DIFF
--- a/sinks/dashboard/items/dashboard_item.py
+++ b/sinks/dashboard/items/dashboard_item.py
@@ -8,6 +8,7 @@ from collections import OrderedDict
 import json
 
 from .no_text_action_parameter import NoTextActionParameter
+from .series_parameter import SeriesChecklistParameter, SeriesListParameter
 
 
 class DashboardItem(QWidget):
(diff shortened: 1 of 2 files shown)


Darth-Raazi (commented):
Tested by running dashboard, adding plot and then running fakeni. That worked. However then I killed fakeni, closed parameter tree and reopened it, and all the FAKE# streams were still showing. Does a refresh pick up that a stream is no longer available?

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/278#-:-O2h2dRMEZZnOsoaaBXG:b67m20v)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @BigLiban, @Lucifersan, and @saads312)

patrick-gu:
No, we cannot detect when a stream is no longer available, at least currently. I think trying to do so would lead to more confusing behaviour.

patrick-gu:
I have tested this with replay log with at least like 20 SDIs and at 10x speed, and it does not appear to have any funny business so I think it works.

Darth-Raazi (approved):
Ok, sounds good
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/278#-:-O2pwoti-D_dGCjtsInt:blfwyhe)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @BigLiban, @Lucifersan, and @saads312)

PR #326, opened 2024-11-11
updated config.py rate and read_bulk to make data more smooth

AshTheEngiqueer:
## Description
Changed RATE and READ_BULK in sources\ni\configuration.py to 20,000 and 2,000 respectively to reduce sensor noise on Flintec load cell

## Developer Testing

Tested new constants with DAQ NI box and saw noise reduced (previously +-200g on flintec load cell, now +- 10g)

## Reviewer Testing

Run dashboard with DAQ sensors and observe noise, compare to old NI config

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/326)


zangjiucheng (approved):
Approve, (Just unblock the enforce git action check, not review the code)

zangjiucheng:
Consider not keep the config.py in the Git Tracking file, since we have ssh server. This can be easy update everytime. We may keep a config.py.example with all comment out.

zangjiucheng:
This PR gonna close, since we are planning to move out the config.py out of the git tracking. An config.py.example will be provided.

Issue #301, opened 2024-10-02
QR Code Generator


zangjiucheng:
- Should offline generate qrcode base on the URL provided. 
- This will like scanner device to the folder we share on the local network. 
- If this qr code can optimize for the Avenza App can be even better. Like pop up the app with auto load the data. 
- QR code should be able to zoom in scale for easy to scan

PR #327, opened 2024-11-12
Qr Code Generating Script

yuvibirdi:
## Description
- Added Feature Request of Generating A QR Code from a url (for ftp/https server)
- Currently there is no input validation on URLs and it will just encode the content in the qrcode, which I don't think is necessary because the url is being generated from another script which should do the validation.
- Moreover, deep links for the Avenza Maps does not work, I talked to the company support for an hour and we determined that is a problem at their own end therefore, currently we have to copy the link from the qr code and import it from the url import section in the app, this will be fixed as soon as I get an update from the company about it.
- For Zoom, the QR Code takes up the maximum size of the Qt window while preserving aspect ratio, on maximizing the window it takes up the full screen which I believe should be efficient enough for it work instead of adding a zoom feature.

- Created new file url_to_code.py containing the code for generating the qr-code and displaying it using pyqt6
- Created ur_to_qrcode.py (currently a test version could be put in .gitignore so it only stays on user machine)
- Added to Config.py  the variable TERMINAL_QR_CODE to print out the qrcode in the terminal in case a GUI environment is not available for backup, also a opening it with a webbrowser has been left as a second backup (but that is commented out) 

This Closes issue #301

## Developer Testing
- Testing with different kinds of links 
- Tested on different screen resolutions for auto scaling
- Set a minimum resolution that should work with all devices out of the box

## Reviewer Testing

- Try Running it on different operating systems (Because I only had the opportunity to test this in on Macos)
- Test and check if the minimum size of the Qr-code work on a multitude of devices
- Test on different monitors with different resolutions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327)


zangjiucheng (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @yuvibirdi)

___
*[a discussion](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327#-OBXNzuYCT6lSFmIBvlN:-OBXNzuYCT6lSFmIBvlO:b-dj61qd) (no related file):*
Excellent! It was implemented very well the first time. Everything looks great, except for one small issue I mentioned in the code block, please fix it! Thanks!

- One more thing to mention, please follow the branch name rule and use {watid}/{issue-description} even in your own folk branch, thanks! (This will make PR checkout for other reviewer easier) 
___
*[`sinks/interamap/src/url_to_qrcode.py` line 65 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327#-OBXNkAlEQXGmUpIaf4z:-OBXNkAlEQXGmUpIaf5-:b459gvk) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/476f7d60bad810e9cbb08d203b6239a58fc9eacf/sinks/interamap/src/url_to_qrcode.py#L65)):*
> ```Python
> 
>         # Determining size wrt, minimums
>         width = max(available_size.width(), self.minimum_width)
> ```

This will cause the widget to only zoom up, not down. Everything else looks good :100: !

yuvibirdi (commented):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @zangjiucheng)

___
*[a discussion](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327#-OBXNzuYCT6lSFmIBvlN:-OBXSB_D02gJoE3S296o:b-5ggm7j) (no related file):*
<details><summary><i>Previously, zangjiucheng (Jiucheng(Oliver)) wrote…</i></summary><blockquote>

Excellent! It was implemented very well the first time. Everything looks great, except for one small issue I mentioned in the code block, please fix it! Thanks!

- One more thing to mention, please follow the branch name rule and use {watid}/{issue-description} even in your own folk branch, thanks! (This will make PR checkout for other reviewer easier)
</blockquote></details>

Got it, how do you want me to change the branch name, because from my understanding I can't do that during a pr is open
___
*[`sinks/interamap/src/url_to_qrcode.py` line 65 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/327#-OBXNkAlEQXGmUpIaf4z:-OBXS8No3rH7irYprWV2:b-lx4fuw) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/476f7d60bad810e9cbb08d203b6239a58fc9eacf/sinks/interamap/src/url_to_qrcode.py#L65)):*
<details><summary><i>Previously, zangjiucheng (Jiucheng(Oliver)) wrote…</i></summary><blockquote>

This will cause the widget to only zoom up, not down. Everything else looks good :100: !
</blockquote></details>

Fixed!

zangjiucheng (approved):
LGTM

PR #344, opened 2025-01-24
Add checkmark for omnibus server launch

zangjiucheng:
## Description
HOTFIX: DAQ issue with launcher should have a checkmark for if omnibus server start on local end. 

- Also a bit formating

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344)

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -13,10 +13,12 @@ from typing import List, Tuple
 
 # Import from Qt – note: these come from pyqtgraph’s wrapper
 from pyqtgraph.Qt import QtGui
+from pyqtgraph.Qt import QtCore
 from pyqtgraph.Qt.QtWidgets import (
     QApplication, QDialog, QLabel, QDialogButtonBox, QVBoxLayout,
     QWidget, QCheckBox, QGridLayout, QLineEdit
 )
+
 
 from logtool import Logger
 
@@ -231,9 +233,9 @@ class Launcher:
             # Poll processes until one finishes or user interrupts
             while True:
                 for process in self.processes:
-                    if process.poll() is not None:
+                    all_finished = all(process.poll() is not None for process in self.processes)
+                    if all_finished:
                         raise Finished
-                time.sleep(0.5)
         except (Finished, KeyboardInterrupt, Exception):
             for process in self.processes:
                 try:
@@ -405,20 +407,22 @@ class GUILauncher(Launcher, QDialog):
         self.selected_ok = True
         # Add omnibus command
 
-        # Construct source commands
-        for i, (selected, args) in enumerate(self.src_state):
-            if selected:
-                cmd = [python_executable, os.path.join("sources", self.modules["sources"][i], "main.py")]
-                if args:
-                    cmd.append(args)
-                self.commands.append(CommandStruct(command=cmd, stdout=False))
+        for idx, src in enumerate(self.src_state):
+            if src[0]:
+                command = [python_executable, f"sources/{self.modules['sources'][idx]}/main.py"]
+                if src[1] != "":
+                    command.append(src[1])
+                source = CommandStruct(command=command, stdout=False)
+                self.commands.append(source)
 
-        # Construct sink commands
-        for i, (selected, stdout_flag) in enumerate(self.sink_state):
-            if selected:
-                cmd = [python_executable, os.path.join("sinks", self.modules["sinks"][i], "main.py")]
-                self.commands.append(CommandStruct(command=cmd, stdout=stdout_flag))
-        self.save_selected_to_config()
+        for idx, snk in enumerate(self.sink_state):
+            if snk[0]:
+                command = [python_executable, f"sinks/{self.modules['sinks'][idx]}/main.py"]
+                if snk[1]: # Check stdout flag
+                    sink = CommandStruct(command=command, stdout=True)
+                else:
+                    sink = CommandStruct(command=command, stdout=False)
+                self.commands.append(sink)
         self.close()
 
     def closeEvent(self, event) -> None:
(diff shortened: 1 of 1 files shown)


ChrisYx511 (changes requested):
Minor behaviour fixes needed, but overall looks good! HOWEVER, some other launcher problems are detailed in a separate comment, which are also important to consider before putting launcher into prod.
TYSM :)

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344#-:-OHQNDI3DAScsucqTLOn:b-77ckg2)* status: all files reviewed, 2 unresolved discussions (waiting on @zangjiucheng)

___
*[`launcher.py` line 282 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344#-OHQMA2O6aXPhxZbRU2W:-OHQMA2O6aXPhxZbRU2X:b-xnw0jv) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ae597820758e0dcf77a64848160d0ed76b82303/launcher.py#L282)):*
> ```Python
>         server_label = QLabel(self)
>         server_label.setText(
>             "Start up with omnibus server? (If you are on DAQ then don't check this)"
> ```

This text is potentially confusing. Maybe change to "Advanced: Start local Omnibus server?  (Do NOT use when the DAQ box is also online)"

Also I'd move the whole checkmark to the bottom of the screen rather than the top.
___
*[`launcher.py` line 287 at r1](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344#-OHQMx4x82Nkz-AlQmlS:-OHQMx4x82Nkz-AlQmlT:bih761h) ([raw file](https://github.com/waterloo-rocketry/omnibus/blob/7ae597820758e0dcf77a64848160d0ed76b82303/launcher.py#L287)):*
> ```Python
> 
>         self.server_checkmark = QCheckBox(self)
>         self.server_checkmark.setChecked(True)
> ```

This should NOT be on by default.

ChrisYx511:
# Issues
## This Patch
- The text next to the checkbox is potentially confusing. See review below.
- Text mode has not been updated, only GUI mode works
## Other Issues [!]
Other Issues that are potentially out of scope but that **WILL** prevent usage of launcher during tests:
- Everything crashes if you have multiple sinks selected and the first sink doesn't find the omnibus server. There is also no way to specify 
- Launcher doesn't work when selecting multiple sinks. When you select both dashboard and globallog for instance, it will just launch 2 dashboards.
- Closing dashboard stops everything? What if I need to keep logging even after closing dashboard?
- What does the stdout option even do? This is just a question?

**I think it might be a good idea we have a discussion about the role of launcher and maybe rethink the architecture.**
TYSM for the patch though! The checkbox does what it needs to do :)

zangjiucheng:
Got it, thanks Chris. I am going to make a discussion about this in the Meeting. Let clarify the behaviour expected from this program.

zangjiucheng:
Hello Chris,

I apologize for the delayed response. I have only recently realized that I have not yet completed the implementation of the requested features.

For this version:
- Refactoring the launcher for improved coding logic
- Enhancing the user interface
- I have been unable to reproduce the issue of launching the omnibus and logger to open two omnibus simultaneously. However, I have identified a potential solution by adding a delay between each service startup and implementing a try-catch block. This may address the issues you mentioned, but please thoroughly test it before finalizing the implementation.
- The closing event will now wait for all threads to complete before exiting.
- stdout will now print output to both the logger and the console, rather than only redirecting to the logger. This will ensure that printer-related information is both logged and displayed on the screen.
- Call Close Function after press `ESC`

launcher.py
```diff
@@ -1,377 +1,444 @@
...
         self.setWindowTitle("Omnibus Launcher")
+        self.setFixedSize(700, 600)
 
-        source_label = QLabel(self)
-        source_label.setText("Sources:")
-        source_label.setGeometry(20, 10, 50, 10)
+        # Checkbox for if start with the omnibus server
+        self.omnibus_checkbox = QCheckBox("Advanced: Start local Omnibus server? (Do NOT use when the DAQ box is also online)")
+        self.omnibus_checkbox.setStyleSheet("font-size: 16px; font-weight: bold;")
+        self.omnibus_checkbox.setChecked(False)
```
ChrisYx511, line 283: Checkbox here is unchecked by default, which is good, but on line 286 the default state of omnibus wasn't updated in consequence. That means omnibus still starts by default unless you toggle the checkbox.

launcher.py
```diff
@@ -1,377 +1,444 @@
...
+        self.omnibus_checkbox.stateChanged.connect(self.on_omnibus_state_changed)
+        
+        self.on_omnibus_state_changed(2)  # default to checked
```
ChrisYx511, line 286: Here is the inconsistent behaviour.

ChrisYx511 (changes requested):
As we discussed, the state of launcher as is will be good enough to use in testing so I'll just request you fix some minor bugs before moving ahead.
- See the comment about the checkbox not working correctly (TL;DR It looks unchecked but still starts omnibus server by default)
- There is no option to not start omnibus server in text mode

Once this is fixed it should be good! All the stuff now does launch properly, including globallog.

launcher.py
zangjiucheng: Wow, nice catching! How I make this mistake....

(I think not-checked should be the default case, so just remove the link 286

ChrisYx511:
LGTM

ChrisYx511 (approved):

ChrisYx511 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344#-:-OJvDUHA5Zmze0Oh8rsd:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @zangjiucheng)

ChrisYx511 (approved):
> *[Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/344)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @zangjiucheng)

zangjiucheng:
Thanks Chris :)

waterloorocketry-common (approved):

PR #349, opened 2025-02-01
Restructure NI Source Config

### LINKED ISSUE #345: [Feature request (NT)] [sources/ni] Move configs out of tree
# Feature Description
- **Describe the feature**
Currently, NI config is simply edited directly in the source tree and saved to the DAQ box, without backups. This feature request seeks to move `config.py` out to the root in a gitignored folder, such as `config/ni.config.py`. Then, all the commented out examples in the current config.py file, documenting the calibration values of the various sensors on the DAQ should be moved into a new file called `ni.config.py.example` and be stored with the NI source source code. 
- **Expected behavior**
  - `sources/ni/config.py` on the DAQ box **SHOULD** be moved to `config/ni.config.py`
  - NI source **MUST** be updated with dynamic import for `config/ni.config.py`, exiting with exception if the file cannot be found.
  - An example configuration file containing the commented out historical sensor calibration data **MUST** to be created at `config/ni.config.py.example`
- **Additional context**
  - If any updates are made to LinearCalibrations or new LinearCalibrations appear during OPS, these should be noted down by designated software member on duty and logged into example config. Ideally do not push changes from DAQ box.
  - DAQ should be periodically plugged into internet and pull the latest example config.

Further Goal: Allows DAQ box to be updated over git / with a zip source package without having to resort to manually copying the config file.

ChrisYx511:
## Description

This PR moves config.py out-of-tree and instead records historical NI configuration in config.example.py. This is to avoid git conflicts when updating the DAQ box.

- Took production `config.py` from DAQ box, cleaned up, formatted using `black`, and renamed to `config.example.py`
- Added conditional import in NI source, exiting with error if config is not found
- Added a lower level .gitignore in `sources/ni` to ignore `/config.py`
- Updated DAQ box with these changes

This PR closes #345 .

## Developer Testing

- Ensured script errors when there is no config.py
- Ensured config.py.example is a valid config
- Testing on DAQ box
- Backed up old DAQ `config.py` in `omnibus-old`

## Reviewer Testing

- Run NI source with a config.py created from config.py.example and ensure it does not crash because of config problems
- Run NI source without config.py and ensure it DOES exit with error

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/349)


sources/ni/config.py.example
zangjiucheng, line 1: Since it comes to .example, please commit out all the lines.(Others make thought that's the line need to uncommitted)  Make this intrusive for our potential user :)

tools/__init__.py
zangjiucheng, line 1: Why do we remove this file? (Don't know if this will cause import error, just confirm)

zangjiucheng (changes requested):
Looks good! Just a few commits, thanks Chris

sources/ni/config.py.example
ChrisYx511: The test values and PORT configuration should not change though? And the rest are all just sane defaults (RATE, READ_BULK, etc.). I think the file should have good default values. Therefore I think they shouldn't be commented. The commented values are only for sensors.

ChrisYx511: Plus, the example file is not getting executed anyways without deliberately making it config.py

tools/__init__.py
ChrisYx511: It was an empty file. Not sure if it had any use.

ChrisYx511: The tools folder is not a module

sources/ni/config.py.example
zangjiucheng: OK, if this could them easier then I think that make sense to work like this :)

tools/__init__.py
zangjiucheng: Yes, it's not a module. I just wanna double check since I don't know if this be designed like thing. If not using sure, we will just delete this file.

zangjiucheng (approved):
LGTM, :+1:

PR #355, opened 2025-02-04
[Core Library] ZMQ Network Recovery after Network Loss

### LINKED ISSUE #302: ZMQ Network Down Detection / Dropped Recovery
- Current our omnibus server rely on the QMZ network, since it can cause any issue went down. We hope to have a detector of network change and when that happen try to reconnect it after a timeout. 
- Currently only solution for us is manual restart 
- (If the auto-detect feature is not available, set up a manual button to try to reconnect to the server.)

ChrisYx511:
## Description

Currently, there is an issue where`Receiver`, which extends `OmnibusCommunicator`, will deadlock when network connection is changed or lost when connecting to an external Omnibus Server. As a result, sinks stop working / fail silently. The solution currently envisioned is to reset the ZMQ socket whenever no messages are being received, so they can recover automatically after connection loss.

- Added new `reset()` function to Receiver class in Omnibus Core library, which re-establishes the connection to the Omnibus server
- Setup the receive message function to run `reset()` after not receiving messages for over 5 seconds (by default, 2 seconds for globallog), and reset every 5 seconds until a new message is received.
- Prints warnings and info logs when network is offline / back online
- Added startup screen incl. build number and date to Omnibus startup
    -  **BREAKING** `gitpython` has been added as a dependency in order to print the build number. It is technically optional and nothing will crash without it, but it should be installed for all operation (so you can get the build number readout)
- Measured performance of reset function, usually can start receiving messages roughly 0.0025 seconds after reset. Therefore maximum theoretical data loss during network failure and reconnect is roughly 5 seconds (2 seconds for globallog) + 0.0025.
- Updated Parsley source to respect this behaviour by listening for Healthcheck as well on the Receiver, not just CAN/Commands. 
- Various other minor fixes, incl. adding a pyproject.toml to replace deprecated `setup.py develop`
- Incremented core library version to `1.1.0`
- Fixed GitHub workflow enforcing version changes

This PR closes #302 .

## Developer Testing

- Tested various disconnection methods, incl. server died, client died, client network changed, client network off using the actual DAQ computer at the bay using all sinks, focusing on Dashboard and Globallog
- Wrote and ran several unit tests that verify message integrity before and after a reset, as well as simulating an auto-reset

## Reviewer Testing

- Run pytest for omnibus, ensure that all tests pass, including old and new ones
- Test disconnecting and reconnecting the server, your client, in various ways you can think of
- Use omnibus normally and report any other regressions as you see fit

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/355)


sinks/dashboard/main.py
```diff
@@ -1,16 +1,31 @@
...
+}
 
 def update():  # gets called every frame
     # read all the messages in the queue and no more (zero timeout)
     while msg := receiver.recv_message(0):
         # updates streams, which them updates the dashitems
         parsers.parse(msg.channel, msg.payload)
+        networkReset["timeout"] = 3
+    if networkReset["lastCheck"] <= 0:
+        networkReset["lastCheck"] = 120
```
zangjiucheng, line 20: For the frame count, can you maybe use the value in config please?

ChrisYx511: I'll be moving this code to the Core Library anyways 🫡

ChrisYx511:
Alright, this should be feature complete!

setup.py
```diff
@@ -9,7 +9,7 @@
 
 setup(
     name='omnibus',
-    version='1.0.0',
+    version='1.0.1',
```
BluCodeGH, line 12: This definitely feels like a 1.1 bump not a 1.0.1 bump. The API is changing and functionality could be considered changing too.

sinks/dashboard/__main__.py
```diff
@@ -0,0 +1 @@
+import main
```
BluCodeGH, line 1: Why'd you add this?

omnibus/pretty_startup.py
BluCodeGH, line 1: I'd really recommend against the core omnibus library printing anything other than errors. Generally you want libraries to assume as little as possible about their environment to be as useful as possible. For example the current HILT code uses curses to do fancy terminal stuff, so this means we couldn't easily integrate omnibus anymore. If you want to add ascii art that's fine, but it should be invoked by the various sources/sinks (and should probably be done in a separate PR than hard networking code changes?)

omnibus/omnibus.py
```diff
@@ -108,39 +111,101 @@ class Receiver(OmnibusCommunicator):
...
-    def __init__(self, *channels):
+    
+    :param *channels: All the channels the receiver should be listening on, pass each as 
+            individual parameters. Only the beginning of the name is matched.
+    :param seconds_until_reconnect_attempt: OPTIONAL - Specify the number of seconds the 
+            Receiver should wait before attempting a reconnection when no messages
+            are being received. Default value is 5 seconds. Keyword parameter only. 
+    """    
+
+    ## PRIVATE PROPERTIES ##
```
BluCodeGH, line 122: In python, properties defined outside of a method are shared by all class instances, pretty sure that isn't what you want here.

omnibus/omnibus.py
```diff
@@ -108,39 +111,101 @@ class Receiver(OmnibusCommunicator):
...
         """
 
         if message := self.recv_message(timeout):
             return message.payload
         return None
+    
+
+    def reset(self):
+        """
+        PRIVATE METHOD
```
BluCodeGH, line 200: Prefix private methods with a single underscore rather than using a comment, eg `_reset()`.

omnibus/omnibus.py
```diff
@@ -108,39 +111,101 @@ class Receiver(OmnibusCommunicator):
...
+    
+    :param *channels: All the channels the receiver should be listening on, pass each as 
+            individual parameters. Only the beginning of the name is matched.
+    :param seconds_until_reconnect_attempt: OPTIONAL - Specify the number of seconds the 
+            Receiver should wait before attempting a reconnection when no messages
+            are being received. Default value is 5 seconds. Keyword parameter only. 
+    """    
+
+    ## PRIVATE PROPERTIES ##
+    __channels: tuple
```
BluCodeGH, line 123: Use a single underscore for private fields, in python the double underscore is reserved for language internals

omnibus/omnibus.py
```diff
@@ -108,39 +111,101 @@ class Receiver(OmnibusCommunicator):
...
 
         If timeout is None this blocks until a message is received. Otherwise it
         waits for timeout milliseconds to receive a message and returns None. A
         zero timeout is supported for nonblocking operation.
+
+        The default timeout is 3000ms.
         """
 
         if self.subscriber.poll(timeout):
+            if self.__disconnected:
```
BluCodeGH, line 168: See my other comments about not printing stuff unless there is an unexpected error. I think adding the `__disconnected` state is kinda complicated and not that useful since we just use it for printing, I'd prefer we print an error when we time out / seem disconnected and otherwise are quiet.

omnibus/omnibus.py
```diff
@@ -108,39 +111,101 @@ class Receiver(OmnibusCommunicator):
...
         self.subscriber = self.context.socket(zmq.SUB)
         self.subscriber.connect(f"tcp://{self.server_ip}:{server.SINK_PORT}")
+        self.__seconds_until_attempt_reconnect = seconds_until_reconnect_attempt
         for channel in channels:
             self.subscriber.setsockopt(zmq.SUBSCRIBE, channel.encode("utf-8"))
+        print(f"[{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] [INFO] Omnibus receiving from {self.server_ip}")
+        
 
-    def recv_message(self, timeout=None):
+    def recv_message(self, timeout=3000):
```
BluCodeGH, line 156: This is a breaking API change. The reconnection functionality shouldn't rely on users calling recv with a timeout, instead you should always internally call `subscriber.poll()` with a short timeout (1 second?) and implement the user-specific timeout manually with a loop.

omnibus/omnibus_test.py
```diff
@@ -110,3 +112,136 @@ def test_timeout(self, monkeypatch):
...
+    # Note that these unit tests are not the only validation of this feature
+    # End-to-end testing must be done to ensure that this also works when
+    # the network environment changes, or the server changes, etc.
+    # This is only an assessment of the reset capabilities of the receiver
+    def test_receive_message_after_server_restart(self, start_stop_capable_server: Callable[[], SpawnProcess], sender: Sender, receiver: Callable[..., Receiver]):
+        server_process: SpawnProcess | None = None
+        try:
+            server_process = start_stop_capable_server()
+            source:Sender = sender()
+            sink: Receiver = receiver("_TESTING")
```
BluCodeGH, line 156: Nit, please make sure the code is autoformatted, not sure why the typing formatting is inconsistent here.

BluCodeGH (changes requested):
The core feature looks good, but I have a handful of thoughts on the surrounding code changes.

omnibus/pretty_startup.py
ChrisYx511: Good point. The main point of this is actually to print the Build Number and Date of the build to facilitate version tracking for both DAQ operator and the DAQ box itself, making bug tracking a whole lot easier. The ASCII art was just for fun LMAO. The other reason it's there specifically is that I wanted it to print when Omnibus server starts too, not just with a sink or a source.
I am realizing that the printing can definitely be a problem in `Receiver` and `Sender` though as those may often be used headlessly.

-> Would the idea of removing the print from `Receiver` and `Sender` but keeping it in the `server()` be friendly to you? And then exporting the print version function as a part of the library so any source or sink that wants to implement it can do so? ASCII art can go too if that's a problem. 

I also don't understand what you mean by "the current HILT code uses curses to do fancy terminal stuff".

tysm!

omnibus/omnibus.py
ChrisYx511: Noted, thanks! Gotta love python 😭

omnibus/omnibus.py
ChrisYx511: Noted 👍

omnibus/omnibus.py
ChrisYx511: Noted 👍

omnibus/omnibus.py
ChrisYx511: This was my original idea too, but I feel like this is unintuitive for the user though.
Let's say the readout looks like this:
image
Now if it just stops printing when it reconnects, I feel like that may be cause for confusion because unless the person using the system reads this exact PR, they won't necessarily know that the behaviour is 1 check every 2 to 5 seconds and then print, especially if the person isn't used to how CLIs stay quiet on success. It also only prints the "Omnibus is back online" message once, (hence the `__disconnected` flag) so I think it's still somewhat quiet?
If that's not acceptable I will make changes accordingly and as required 🫡.

omnibus/omnibus_test.py
ChrisYx511: Noted 👍

sinks/dashboard/__main__.py
ChrisYx511: To run `python sinks/dashboard` instead of `python sinks/dashboard/main.py`. Will probably revert it for now it was just for convenience.

omnibus/omnibus.py
ChrisYx511: Are you suggesting something like this?
```python
start_time = time.time()
time = start_time
while timeout == None or time - start_time <= timeout / 1000: # I'll figure out a better way of doing the math
    data = self.subscriber.poll(0) # this should be non blocking?? im not sure...
    if data:
    
        return ... # return the message here...
    """ Check for the refresh update here """
    time = time.time()

return None
```
Alternative implementation:
```python
# Non-Blocking Operation
if timeout == 0:
    # do the current implementation
    if msg:=self.subscriber.poll(0)
        return Message....
    """ Check for the refresh update here (now - self.last_updated....) """
    return None
# Blocking operation
i = 0
while timeout == None or i <= timeout: # I'll figure out a better way of doing the math
    data = self.subscriber.poll(1) # this will essentially execute 1ms * timeout times, making a total of timeout ms
    if data:
        self.last_updated = time.time()
        return.... # return the message here...
    """ Check for the refresh update here (now - self.last_updated....) """"
    if (timeout):
        i += 1 # Prevent overflow if no timeout is set
return None
```

ChrisYx511: If this is what you're thinking of, are any of the implementations good? / Which one do you think is better?

ChrisYx511:
@BluCodeGH Thank you so much for the insight! I just had some quick clarifying questions that I left as comments to your review and I'll get right to implementing the changes once they are answered.
Thanks again!

omnibus/pretty_startup.py
BluCodeGH: Yep I think only having the print in server and exposing the version number as a getter makes sense. HILT is the hardware in the loop tester we use on electrical, right now its unrelated to omnibus but in the future we might want to integrate functionality. Code is here if you're curious: https://github.com/waterloo-rocketry/warrior/pull/1

omnibus/omnibus.py
BluCodeGH: Hmm yeah I understand, I guess a flag is the best option.

omnibus/omnibus.py
BluCodeGH: Yes this is along the lines of what I'm thinking of. You should still do a blocking poll since nonblocking will eat 100% cpu when called in a loop like that. You can probably do something closer to the first implementation but block for `min(1, requested_timeout)`.

zangjiucheng:
Thanks @BluCodeGH. Since this is the core library of omnibus, I believe your idea is more insight lol. I will keep quite before your ideas be addressed. I will just do last check after your review and send an approve. Thanks again

ChrisYx511:
@BluCodeGH Ready for re-review whenever you're available 👍

BluCodeGH (changes requested):
Looking good, more of a nitpicky pass this time.

omnibus/omnibus.py
```diff
@@ -76,8 +85,26 @@ class Sender(OmnibusCommunicator):
 
     def __init__(self):
         super().__init__()
-        self.publisher = self.context.socket(zmq.PUB)
-        self.publisher.connect(f"tcp://{self.server_ip}:{server.SOURCE_PORT}")
+        try:
+            assert (
+                OmnibusCommunicator.context != None
+                and OmnibusCommunicator.server_ip != None
+            )
```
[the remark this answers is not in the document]
ChrisYx511: Reason 1:
The idea that I discussed with Oliver & Kavin was that we wanted to gradually introduce strict type checking into the Omnibus repo, in order to improve readability, code completion, and reduce potential bugs. Since both context and server_ip are defined as `[Something] | None` type, we have to either `cast` or `assert` to respect the rule of typechecking despite knowing that "technically" it will never be `None` at that point. In accordance with what I've read about best practices when it comes to Pyright checks, we should use an `assert` rather than a `cast`. 
Ref. https://github.com/microsoft/pyright/issues/7706#issuecomment-2057222681

Reason 2:
Also what if a faulty library user tries to set it to None for some reason? Now that the socket might restart during the lifetime of the application, we need to ensure that the context and IP are always available so we can reset the socket as needed. If you never trigger the reconnect condition you will never detect the faulty behaviour. While you can argue that it's up to the user to not do stupid things with the library, Python doesn't exactly enforce that behaviour and I thought failing explicitly might be nicer than failing kinda randomly when the library tries to reconnect. It might just look like random crashes otherwise.

Let me know what you think! And tbf the try except block might be excessive I can look into that.

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
+    ## PRIVATE PROPERTIES ##
+    _channels: tuple[str, ...]
+    # Keep track of last received message, only second granularity is needed
+    # so time.time() is good enough on any platform
+    _last_online_check: float
+    _disconnected: bool
+    _seconds_until_attempt_reconnect: int
+    ## END PRIVATE PROPERTIES ##
+
+    subscriber: zmq.SyncSocket | None = None
```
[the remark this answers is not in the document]
ChrisYx511: In accordance with PEP 526 (https://peps.python.org/pep-0526/), integrated into standard in 2016: 
"the value-less notation a: int allows one to annotate **instance variables** that should be initialized in `__init__` or `__new__`".

If a variable is a class member, it should be annotated with `ClassVar[...]` (https://typing.readthedocs.io/en/latest/spec/class-compat.html#classvar). This is also in an effort to move towards static type annotation as that really helps readability and just making working with Omnibus nicer.

Since we are targeting Python 3.11, which includes this PEP in the standard, my interpretation of the above documentation leads me to believe that this is actually the right way to do this and to further improve readability as you mentioned. Example from the docs:

```python
class Starship:
    captain: str = 'Picard'
    damage: int
    stats: ClassVar[dict[str, int]] = {}

    def __init__(self, damage: int, captain: str = None):
        self.damage = damage
        if captain:
            self.captain = captain  # Else keep the default

    def hit(self):
        Starship.stats['hits'] = Starship.stats.get('hits', 0) + 1

enterprise_d = Starship(3000)
enterprise_d.stats = {} # Flagged as error by a type checker
Starship.stats = {} # This is OK
```
I'm not sure how I am deviating from Python standards? But let me know if I'm completely wrong LOL.

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
+        """
         super().__init__()
+        self._last_online_check = time.time()
+        self._channels = channels
+        self._seconds_until_attempt_reconnect = seconds_until_reconnect_attempt
+        self._disconnected = False
+        self._connect()
+        print(
+            f"[{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] [INFO] Omnibus receiving from {OmnibusCommunicator.server_ip}"
+        )
```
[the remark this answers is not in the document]
ChrisYx511: Noted, will fix 🫡

omnibus/omnibus.py
```diff
@@ -144,3 +298,14 @@ def recv(self, timeout=None):
...
+
+    def _reset(self):
+        """
+        Restart the ZMQ Socket of this receiver instance. Called when the Receiver
+        has not received any data for over a given number of seconds, specified
+        by seconds_until_reconnect_attempt.
+        """
+        if self.subscriber != None:
+            self.subscriber.close(linger=0)
+            self.subscriber = None  # Prevent receiver from attempting to read from socket while it's closed.
```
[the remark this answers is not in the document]
ChrisYx511: OK, I'll remove the None state from subscriber.

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
         zero timeout is supported for nonblocking operation.
         """
-
-        if self.subscriber.poll(timeout):
-            channel, timestamp, payload = self.subscriber.recv_multipart()
-            return Message(channel.decode("utf-8"), msgpack.unpackb(timestamp), msgpack.unpackb(payload))
+        try:
+            assert (
+                self.subscriber != None
+            )  # Ensure subscriber is actually ready to receive, should never trigger in theory
```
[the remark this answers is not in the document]
ChrisYx511: Will remove 👍

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
+            actual_timeout = (  # Ensure that we do still check for network loss even when timeout is large or infinite
+                self._seconds_until_attempt_reconnect * 1000
+                if timeout == None
+                else min(timeout, self._seconds_until_attempt_reconnect * 1000)
+            )
+            i = 0
```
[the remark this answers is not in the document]
ChrisYx511: Noted 👍

omnibus/omnibus.py
```diff
@@ -86,11 +113,21 @@ def send_message(self, message: Message):
         Note that channel used is specified by the message object rather than
         the sender.
         """
-        self.publisher.send_multipart([
-            message.channel.encode("utf-8"),
-            msgpack.packb(message.timestamp),
-            msgpack.packb(message.payload)
-        ])
+        try:
+            assert self.publisher != None
```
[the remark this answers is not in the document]
ChrisYx511: Will remove 👍

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
+
+    def _connect(self):
+        """
+        Create ZMQ subscriber and connect to server
+        """
+        try:
+            assert (
+                OmnibusCommunicator.server_ip != None
+                and OmnibusCommunicator.context != None
+            )
```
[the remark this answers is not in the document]
ChrisYx511: See above for discussion

omnibus/omnibus.py
```diff
@@ -108,31 +145,148 @@ class Receiver(OmnibusCommunicator):
...
+            while timeout == None or i < timeout or timeout == 0:
+                poll_result = self.subscriber.poll(timeout=actual_timeout)
+                self._check_online_and_reconnect(
+                    poll_result
+                )  # Will attempt to reconnect here if reconnection timeout exceeded
+                if poll_result == 0:  # No message received
+                    if timeout == 0:
+                        break
+                    if timeout != None:
+                        i += actual_timeout  # + 1 to prevent being stuck in the loop if timeout == 0
```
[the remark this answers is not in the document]
ChrisYx511: oops

omnibus/server.py
```diff
@@ -40,21 +45,31 @@ def ip_broadcast():
...
+    )  # Initialize BuildInfoManager to print build info
+    bim.print_startup_screen()
+    bim.print_app_name()
+
     context = zmq.Context()
 
-    proxy = ThreadProxy(zmq.SUB, zmq.PUB, zmq.PUB)  # proxies messages in a separate thread
+    proxy = ThreadProxy(
+        zmq.SUB, zmq.PUB, zmq.PUB
+    )  # proxies messages in a separate thread
```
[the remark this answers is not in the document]
ChrisYx511: I don't see an autoformatter, only like format checkers, in format.sh. I just ran `black` with default settings. But yeah I'll reformat with 100 chars as defined for the enforcer in flake8. Eventually we should also add an autoformatter to the pipeline.

sinks/dashboard/main.py
```diff
@@ -8,9 +8,10 @@
 
 def update():  # gets called every frame
     # read all the messages in the queue and no more (zero timeout)
+    print("update")
```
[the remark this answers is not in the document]
ChrisYx511: oops x2

sources/parsley/main.py
```diff
@@ -140,24 +141,25 @@ def main():
...
-                    formatted_msg += ',' + ','.join(f"{byte:02X}" for byte in msg_data)
-                formatted_msg += ";" + crc8.crc8(
-                    msg_sid.to_bytes(4, byteorder='big') + bytes(msg_data)
-                ).hexdigest().upper()  # Sent messages in the usb debug format have a crc8 checksum at the end, to be investigated: https://github.com/waterloo-rocketry/omnibus/commit/0913ff2ef1c38c3ae715ad87c805d071c1ce2c38
-                print(formatted_msg)  # Always print the usb debug style can message
-                # Send the can message over the specified port
-                communicator.write(formatted_msg.encode())
-                last_keepalive_time = now
-                time.sleep(0.01)
+            if msg.channel.startswith(RECEIVE_CHANNEL):
```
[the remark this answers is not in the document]
ChrisYx511: Noted!

omnibus/omnibus.py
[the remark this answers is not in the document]
ChrisYx511: In fact if anything we should annotate context and server_ip with ClassVar.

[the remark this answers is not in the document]
BluCodeGH: Oh my mistake, TIL. In that case I'd just remove the `PRIVATE PROPERTIES` comments since that's implicit from the underscores.

omnibus/omnibus.py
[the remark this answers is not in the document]
ChrisYx511: Final Decision: Remove try except, keep assert due to more semantic errr

omnibus/server.py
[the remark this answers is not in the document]
ChrisYx511: The official suggestion is 88 characters, I might just move the comment up and stick with that.

ChrisYx511:
@BluCodeGH @zangjiucheng pls re-review

zangjiucheng (commented):
LGTM, thanks Chris that's really amazing... I will even say it's 2.0 instead of 1.1 for core library lol. I leave a few comments there. 

Code makes sense to me, I will test on that with usb debugger tmr to confirm it keeps working with real parsley signal. As soon as tomorrow an approve of this pr will be send.

omnibus/omnibus.py
```diff
@@ -108,31 +128,136 @@ class Receiver(OmnibusCommunicator):
...
+            self._last_online_check = time.time()
+            # If we were disconnected, print that we are now reconnected
+            if self._disconnected:
+                print(
+                    f"[{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}] [INFO] Omnibus Receiver is online!"
+                )
+            self._disconnected = False
+            return
+        # No message received
+        if (
```
zangjiucheng, line 205: Maybe here need a initial state, since just start the DAQ, it may don't have the message. This should be only trigger after received on the first message. All state come online.

zangjiucheng:
image


For feature improvement, 

1. for the point when server got disconnected, maybe we need a better identification for plot. 
2. Also dashboard title maybe can give a UI reflection say server is down.
3. Make sure in that case sender is not be able to send any signal (Maybe we should stash command need to be send in the buffer area when server is down)

zangjiucheng:
Please create relative issues before we merge this in (Make sure we won't forgot those features wait to be fixed) 

(Could be just a title, I am going to give detail description later)

omnibus/omnibus_test.py
```diff
@@ -110,3 +137,253 @@ def test_timeout(self, monkeypatch):
...
+            server_process = start_stop_capable_server()
+            source: Sender = sender()
+            sink: Receiver = receiver("_TESTING")
+            source.send("_TESTING", "PAYLOAD")
+            assert sink.recv(10) == "PAYLOAD"
+            server_process.terminate()
+            server_process.join()
+            server_process = None
+            server_process = start_stop_capable_server()
+            time.sleep(0.5)
```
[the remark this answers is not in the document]
ChrisYx511: I believe so, as we are terminating the server, then restarting, all in a separate thread. The sleep gives it time to come online before we test if we can connect to it.

omnibus/omnibus.py
```diff
@@ -108,31 +128,136 @@ class Receiver(OmnibusCommunicator):
...
+    _channels: tuple[str, ...]
+    # Keep track of last received message, only second granularity is needed
+    # so time.time() is good enough on any platform
+    _last_online_check: float
+    _disconnected: bool
+    _seconds_until_attempt_reconnect: int
+
+    _subscriber: zmq.SyncSocket
+
+    def __init__(self, *channels: str, seconds_until_reconnect_attempt: int = 5):
```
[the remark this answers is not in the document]
ChrisYx511: It's set to 2 for globallog, 5 for rest seems reasonable just not to send too many reconnects all the time. What do you think?

zangjiucheng (approved):
LGTM, after testing on the GPS parser after message go offline. It can back online as expected.

omnibus/omnibus.py
ChrisYx511: Disagree, I think that if it's disconnected from the beginning it should still report it. Let's say the server dies for some reason, you wouldn't know unless it warns you.

ChrisYx511: Also if you look at init() it already starts in not disconnected state so I'm not really sure?? 
Let me know if I'm misinterpreting and I can fix it.

ChrisYx511:
@zangjiucheng Did a minor typing fix, could you re-approve? 🙏 Also left a comment https://github.com/waterloo-rocketry/omnibus/pull/355#discussion_r1979871079.

omnibus/omnibus.py
zangjiucheng: Let me have a check and get you back soon

omnibus/omnibus.py
```diff
@@ -108,31 +128,136 @@ class Receiver(OmnibusCommunicator):
...
+                    # but if you have 5+ second timeouts I don't think 1ms matters
+                    actual_timeout = min(timeout - elapsed_time_ms, actual_timeout)
+                continue
 
-        if self.subscriber.poll(timeout):
-            channel, timestamp, payload = self.subscriber.recv_multipart()
-            return Message(channel.decode("utf-8"), msgpack.unpackb(timestamp), msgpack.unpackb(payload))
+            # If there is a message received, proceed below:
+            channel, timestamp, payload = self._subscriber.recv_multipart()
+            self._last_online_check = time.time()
```
BluCodeGH, line 252: Nit, you already set this in `_check_online_and_reconnect()` no?

zangjiucheng (approved):
LGTM

zangjiucheng (approved):

omnibus/omnibus.py
ChrisYx511: Fixed.

zangjiucheng (approved):

PR #356, opened 2025-02-05
Add Gradient Colors

vincentjguo:
## Description
Added gradient colors to linestring
Configurable by config

## Developer Testing

- 👀 
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/356)

### DIFF
--- a/sinks/interamap/src/map_view.py
+++ b/sinks/interamap/src/map_view.py
@@ -7,7 +7,7 @@ import random
 from PySide6.QtWebEngineWidgets import QWebEngineView
 from PySide6.QtWidgets import QSizePolicy
 
-from config import ONLINE_MODE, ZOOM_MAX, ZOOM_MIN
+from config import ONLINE_MODE, ZOOM_MAX, ZOOM_MIN, GRADIENT_COLORS
 from src.gps_cache import GPS_Cache
 from PySide6.QtCore import Signal
 import flask
@@ -24,7 +24,7 @@ if not ONLINE_MODE:
 
 import folium
 from folium.plugins import Realtime
-from fastkml import kml
+import branca.colormap as cm
 
 from src.kmz_parser import KMZParser
 from src.data_struct import Point_GPS, LineString_GPS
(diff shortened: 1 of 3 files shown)


sinks/interamap/src/map_view.py
```diff
@@ -249,9 +250,17 @@ def draw_gps_data(self, points: List[Point_GPS | LineString_GPS] = None):
...
                 ).add_to(self.m)
             elif isinstance(data, LineString_GPS):
-                line = folium.PolyLine(
-                    locations=[[point.lat, point.lon] for point in data.points],
-                    color="red",
+                points, color_idx = [], []
+                for point in data.points:
+                    points.append([point.lat, point.lon])
+                    color_idx.append(self.gradient_count)
+                    self.gradient_count = (self.gradient_count + 1) % len(GRADIENT_COLORS)
```
zangjiucheng, line 257: Sorry I didn’t make it clear last time, but I think using a different colour for the most recent line would make it easier for users to track the latest changes to the rocket tracking data. (For any other string just with a default color)

zangjiucheng (changes requested):
Nice Working! It's pretty fast to turn around for this PR! Thanks Vincent

Only a tiny behavior need to be change state in the code comment (Sorry I didn't state clear last time) 

I get a quick question for here, is that possible to support string parse with real-time parser. How hard it could be? (I mean only after click stop parsing) This could be in the next PR. (I believe this require deal with `real_time_parser.py` a bit....

sinks/interamap/src/map_view.py
vincentjguo: That's... very different. It would require redrawing the last point but I don't think folium allows undrawing a single point which might require the entire map to be redrawn 
Maybe we need to move this to another issue so that the requirements are fully defined

zangjiucheng: Sure, we can move that to another issue, then let's merge this now :)

zangjiucheng: But, yes, my idea for it is just redrawing the whole map.... I believe that's the only way, because Folium is not really doing well with dynamic refresh. (Also we need to optimize the point plot, example sample point from few points)

zangjiucheng (approved):

PR #361, opened 2025-02-25
Added Low Pass Filter feature

### LINKED ISSUE #336: [Feature request (NT)] - Low-Pass Filter for Dashboard Plot (Dashboard Sink)
## Feature Description
The dashboard currently displays plots generated from various source messages. However, noisy data often passes through, causing unexpected graph scaling and reduced clarity. To address this, a Low-Pass Filter feature will be implemented.

## Expected Behavior
- A **Low-Pass Filter** option will be available in the **Settings Panel** on the right side of the dashboard.
- Users can configure the filter to suppress values that exceed or fall below the specified threshold, effectively reducing noise in the graph.
- The graph will update dynamically based on the filter settings, ensuring a clean and accurate representation of the data.
image

### Key Features
- [x] **Low-Pass Filter Toggle**: Enable or disable the filter functionality.
- [x] **Threshold Configuration**: Input fields or sliders to specify upper and lower threshold values for the filter.
- [x] **Dynamic Plot Update**: The plot will automatically adjust to display only data within the configured range.

## Additional Context
This feature is intended to enhance user experience by:
- Improving graph readability.
- Mitigating the impact of outlier data points.
- Allowing users to focus on the most relevant data within their desired range.

ChrisWu0000:
## Description
I added a low-pass filter feature to the Plot_Dash_Item, allowing users to filter out data points that exceed a configurable min/max threshold. This improves data visualization by removing outliers and unwanted noise.

This PR closes #336.

## Developer Testing

- Enabled and disabled the low_pass_filter toggle to see if it affects the graph.
- Set different min_threshold and max_threshold values and verified correct filtering.
- Confirmed that out-of-range data points are ignored while valid points are plotted.
- Checked dynamic updates, changing thresholds and graph should immediately affect incoming data.
- If min_threshold > max_threshold, the values are swapped automatically.
- When the filter is off, all points should be displayed as expected.

## Reviewer Testing

- Toggle the Low-Pass Filter ON/OFF and check if filtering applies correctly.
- Set Min and Max Threshold values and verify that out-of-range data is removed.
- Ensure that disabling the filter restores all data points.
- Check UI behavior when modifying threshold values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/361)

### DIFF
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -30,7 +30,10 @@ class PlotDashItem(DashboardItem):
 
         self.parameters.param('series').sigValueChanged.connect(self.on_series_change)
         self.parameters.param('offset').sigValueChanged.connect(self.on_offset_change)
-
+        self.parameters.param('low_pass_filter').sigValueChanged.connect(self.on_low_pass_filter_toggle)
+        self.parameters.param('max_threshold').sigValueChanged.connect(self.on_threshold_change)
+        self.parameters.param('min_threshold').sigValueChanged.connect(self.on_threshold_change)
+        
         self.series = self.parameters.param('series').value()
         # just a single global offset for now
         self.offset = self.parameters.param('offset').value()
@@ -56,7 +59,10 @@ class PlotDashItem(DashboardItem):
         series_param = SeriesChecklistParameter()
         limit_param = {'name': 'limit', 'type': 'float', 'value': 0}
         offset_param = {'name': 'offset', 'type': 'float', 'value': 0}
-        return [series_param, limit_param, offset_param]
+        low_pass_filter_toggle = {'name': 'low_pass_filter', 'type': 'bool', 'value': False}
+        low_pass_filter_max = {'name': 'max_threshold', 'type': 'float', 'value': float('inf'), 'visible': False}
+        low_pass_filter_min = {'name': 'min_threshold', 'type': 'float', 'value': float('-inf'), 'visible': False}
+        return [series_param, limit_param, offset_param, low_pass_filter_toggle, low_pass_filter_max, low_pass_filter_min]
 
     def on_series_change(self, param, value):
         if len(value) > 6:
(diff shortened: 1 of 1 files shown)


ChrisYx511:
Hi! This seems about right from what I understand, but I would consult with the electrical people to make sure that this is what they meant. You can ping @ Ash Lang on Slack in #electrical (or I can do it if you prefer) and we can talk more during meeting next week! I / @zangjiucheng  will approve then.
Thanks for the PR!

ChrisWu0000:
Could you ping them for me? Thanks

zangjiucheng (changes requested):
Screenshot 2025-03-04 at 12 28 30 AM

Nice try!! the function is what we wished. 

There is a major bug issue with this low pass filter now, if the value is not in the filter area, the plot will stop updating, We want it be stream even it don't have any value.

zangjiucheng:
Yes, that's expected behaviour thanks for confirming.

zangjiucheng:
Also please update the branch before we merge this in. Thanks

ChrisWu0000:
I think I fixed the bug

zangjiucheng:
Nice work! Code makes sense to me... let me have a test ton some datapoint tmr. Should be ready by end of tomorrow.

zangjiucheng (changes requested):
Few more changes required here: 

1. After reloading saved dashboard the low_pass_filter is on but parameter box should also display
image

2. After using low_pass_filter, when we disabled, data point is not be restored. When you do the toggle off you should make sure it also refresh the plot 
Screenshot 2025-03-05 at 10 44 50 PM
﻿
3. After reset the value to -inf or inf you will need to make sure it's not be treated as a string.  
Screenshot 2025-03-05 at 10 43 24 PM

4. Please update the PR to the latest, since we get some core library update.

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -105,7 +125,11 @@ def on_data_update(self, stream, payload):
         time, point = payload
 
         point += self.offset
-
+        if self.parameters.param('low_pass_filter').value(): #checks if value is within threashold
+            max_threshold = self.parameters.param('max_threshold').value()
+            min_threshold = self.parameters.param('min_threshold').value()
+            if point > max_threshold or point < min_threshold:
+                point = np.nan
```
zangjiucheng, line 132: Maybe you should set it to the max_threshold or min_threshould instead to the np.nan. which should show a continue lines inside the break graph.

ChrisYx511:
Unfortunately, the feature does not match what is expected from a software low pass filter. The actual implementation would require a lot more work and might not yield much better results than what we currently have, which is just averaging. (Essentially the idea was to actually smooth out the noise rather than cut it out, but we already sort of do that with averaging so electrical is just thinking of doing more hardware work to make it less noisy.)
@zangjiucheng @ChrisWu0000 if that's OK with you guys, I would like to move to de-scope for now. Sorry for the inconvenience.

ChrisWu0000:
Yeah thats fine with me, thanks for telling us.

ChrisYx511:
@ChrisWu0000 If you want something else to work on msg me on Slack, I have interesting work for you!

PR #365, opened 2025-03-07
[sources/ni] CONFIG PATCH - Add PSE Sensor to config.example.py 

ChrisYx511:
Literally just a comment change. Will be used at LCF3.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/365)


ChrisYx511:
Descoped for now.

PR #367, opened 2025-03-07
[sinks/dashboard] HOTFIX - IMPORTANT: Set Plot Title to 100 char max instead of 50 + Break into 2 lines 

### LINKED ISSUE #366: [BUG] - Plot Title not filling available space
# Bug Report
- **Describe the bug**
> Title of Plot dash item does not take up the available space - even when there is more than sufficient space the title will remain the same size and will truncate the title if it is too long

- **To Reproduce**

    > 1. Launch Dashboard
    > 2. Create a plot item for a series with a sufficiently long name (~25 chars) - ex. FPT-101
    > 3. Keep PT disconnected (to get some non-zero values)
    > 4. Title is cut off

- **Expected behavior**
> Expect that title remains centered and shows full string if sufficient space

- **Screenshots**
> 

Image

> Title should show "current: [VALUE] [FPT-101] Fuel Tank (psi)"

- **Environment:**
   - OS Version: Windows 11 Home
   - OS build: 22631.4890
   - CPU Architecture: x86
   - Python Version: 3.12

- **Additional context**

ChrisYx511:
## Description

Currently, plot title is limited to 50 characters and the current values are all on one line. Reverts change from @BluCodeGH. :(

- Set limit to 100 chars
- Added HTML formatting to break current value and series names into 2 lines

This PR closes #366 .

## Developer Testing

- Ran dashboard, title displays properly

## Reviewer Testing

- Ran dashboard, ensure long plot title displays properly

### DIFF
--- a/sinks/dashboard/items/plot_dash_item.py
+++ b/sinks/dashboard/items/plot_dash_item.py
@@ -1,5 +1,5 @@
 from publisher import publisher
-from pyqtgraph.Qt.QtWidgets import QGridLayout, QMenu
+from pyqtgraph.Qt.QtWidgets import QGridLayout, QMenu, QLabel
 from pyqtgraph.parametertree.parameterTypes import ChecklistParameter
 from pyqtgraph.Qt.QtCore import QEvent
 import pyqtgraph as pg
@@ -40,8 +40,8 @@ class PlotDashItem(DashboardItem):
             publisher.subscribe(series, self.on_data_update)
 
         # a default color list for plotting multiple curves
-        # yellow green cyan white blue magenta
-        self.color = ['k', 'g', 'c', 'w', 'b', 'm']
+        # black blue magenta green cyan white
+        self.color = ['k', 'b', 'm', 'g', 'c', 'w']
 
         # create the plot
         self.plot = self.create_plot()
@@ -152,22 +152,25 @@ class PlotDashItem(DashboardItem):
         self.plot.setXRange(t - config.GRAPH_DURATION + config.GRAPH_STEP,
                             t + config.GRAPH_STEP, padding=0)
 
+        series_name: str  = ""
+        # data series name
+        series_name += " / ".join(self.series) # join all data series on plot
+
         # value readout in the title for at most 2 series
-        title = ""
+        current_values: str = ""
         if len(self.series) <= 2:
             # avg values
-            title += "    current: "
+            current_values += "Current: "
             last_values = [self.points[item][-1]
                            if self.points[item] else 0 for item in self.series]
             for v in last_values:
-                title += f"[{v: < 4.4f}]"
-            title += "    "
-        # data series name
-        title += "/".join(self.series)
-        if len(title) > 50:
-            title = title[:50]
+                current_values += f"[{v: < 4.4f}] "
 
-        self.plot.setTitle(title)
+        # 100 CHARS MAX for title
+        series_name = series_name[:100]
+        current_values = current_values[:100]
+
+        self.plot.setTitle(title=f"<div style='text-align: center; width: 100%; font-size: 12pt; padding-top: 100px;'><br/><b>{series_name}</b><br/>{current_values}</div>")
 
     @staticmethod
     def get_name():


sinks/dashboard/items/plot_dash_item.py
```diff
@@ -40,8 +40,8 @@ def __init__(self, *args):
             publisher.subscribe(series, self.on_data_update)
 
         # a default color list for plotting multiple curves
-        # yellow green cyan white blue magenta
-        self.color = ['k', 'g', 'c', 'w', 'b', 'm']
+        # black blue magenta green cyan white
```
zangjiucheng, line 40: Maybe we really need to consider to switch to use Hex color instead. (Make some high contract color) 

 I fell Black + Blue is really notable under the sunshine

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -152,22 +149,25 @@ def on_data_update(self, stream, payload):
...
-                title += f"[{v: < 4.4f}]"
-            title += "    "
-        # data series name
-        title += "/".join(self.series)
-        if len(title) > 50:
-            title = title[:50]
+                current_values += f"[{v: < 4.4f}] "
+
+        # 100 CHARS MAX for title
+        series_name = series_name[:100]
```
zangjiucheng, line 167: image

For extreme case 100 is not a good number lol, maybe a dynamic font size is useful.

image

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -152,22 +149,25 @@ def on_data_update(self, stream, payload):
...
             last_values = [self.points[item][-1]
                            if self.points[item] else 0 for item in self.series]
             for v in last_values:
-                title += f"[{v: < 4.4f}]"
-            title += "    "
-        # data series name
-        title += "/".join(self.series)
-        if len(title) > 50:
-            title = title[:50]
+                current_values += f"[{v: < 4.4f}] "
```
zangjiucheng, line 164: How this 4.4 magic number come from our plot, can we just remove it or set a constant here?

zangjiucheng (changes requested):
Nice work!! Few comments address in the code.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: I believe these colours are fixed by pyqtplot, and we never should NEVER use more than 2 series per plot in production anyways. Green and Cyan were hard to see so I swapped their order. I kept them in the array because otherwise it crashes when more series are selected. I think it's out of scope to fix this for now since this is a hotfix.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: This is not a float. It formats to 4 decimal points with at least 4 chars, and left aligns the result.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: I was testing at the bay and they were a lot more readable. Black is also perfectly fine.

sinks/dashboard/items/plot_dash_item.py
zangjiucheng: Oh thanks. My bad.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: Again, because this is hotfix, I really need this today and I don't want to put too many features, so it's easy to merge.

sinks/dashboard/items/plot_dash_item.py
zangjiucheng: Fine just like we can fix one point each time, this should be flagged. Because I tested it under sunshine it's really hard to check it clear. This software most of time will be used outside. 

Can you revert this change for a second, we can come back to craft something for in feature PR

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: No, because it got cut off.
It used to be like : current: [4.0233] [NPT-201] Nitrogen Fill (ps <- it get's cut off here
Now that it's on 2 lines 50 chars might be fine but just in case I'm making it 100 chars, which is more than reasonable for ~3 series names. In production it is protocol to use MAX 2 anyways.
Like 
[NPT-201] Nitrogen Fill (psi) / [OPT-102] Ox Fill (psi)
This is 55 characters already. 100 is fine.

zangjiucheng: Well, that make sense. Then just last comment here. If the text be cut off please show [char(100)...] with "..." which means not finished. Thanks!

ChrisYx511: Alright, will fix like right now.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: Will do. Thanks!

ChrisYx511: I'll keep white at the end tho because it's really useless. (You just can't see white).

ChrisYx511: I genuinely think the blue is a lot more visible than the green though that's why I want it to be second in the list. I just changed the order around. It's not perfect but it's better and a really simple change. Can we still make the change?

ChrisYx511: @zangjiucheng

zangjiucheng: Ok, since your are also the user of it lol. Please keep it

zangjiucheng (approved):
LGTM, Nice work!

ChrisYx511:
> LGTM, Nice work!

Thank you!

PR #370, opened 2025-03-09
Add CAN Logger board log file parse support

JasonBrave:
## Description

This PR closes #XXX.

## Developer Testing

- Ran A
- Ran B
- Ran C

## Reviewer Testing

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/370)

### DIFF
--- a/sources/parsley/main.py
+++ b/sources/parsley/main.py
@@ -40,7 +40,7 @@ class FileCommunicator:
         return self.file.read(4096)
 
     def write(self, msg: bytes):
-        print('cannot write to a file')
+        print('Cannot write to file: {msg}')
 
 # Acting as a fake usb debug board
 class FakeSerialCommunicator:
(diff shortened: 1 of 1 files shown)


sources/parsley/main.py
```diff
@@ -32,6 +32,15 @@ def read(self):
...
 
+class FileCommunicator:
+    def __init__(self, filename: str):
+        self.file = open(filename, mode='rb')
+
+    def read(self):
+        return self.file.read(4096)
+
+    def write(self, msg: bytes):
+        print('cannot write to a file')
```
ChrisYx511, line 43: Maybe have this as the same format as the Fake Communicator? Something like:
`"Cannot write to file: {msg}"`

ChrisYx511 (commented):
Nits: Split the communicator classes into a separate `communicator.py`? And maybe a `Communicator` interface that declares the common methods among all three types of communicators?
Otherwise looks good, these are just nits. See below for more small nits.

ChrisYx511 (changes requested):
Also update the submodule to point to latest commit of Parsley once parsley commit merged.

ChrisYx511 (approved):
LGTM

PR #378, opened 2025-04-01
Added line of best fit function and display

### LINKED ISSUE #374: [Feature request (NT)] - [sinks/dashbaord] Display Slope of Best Fit Line in Plot
# Feature Description
- **Describe the feature**
A toggle in the property tree of the plot dashboard element should enable the display the numerical value of the slope of the best-fit line across the latest n number of points, by default 10.
- **Expected behavior**
    - When Slope is enabled, should display slope as an extra line in title and update with the plot (i.e. "Slope: [1.03] [1.30]" etc.) 
    - Calculate slope from user specified number of latest points (default 10)
    - The option should available only when number of series in plot less than or equal to 2
    - Slope should be hidden when more than 3 series are selected, and the checkbox unticked and disabled
- **Additional context**
You can probably just use numpy polyfit, get the slope.
Something like this:
Image

Complete by Mar 27.

ChrisWu0000:
## Description
I added a line of best fit feature to the Plot_Dash_Item, which gives you the value of the slope of the line when toggled. I decided to make it so you can still edit the number of data points even if 'Show Slope' is disabled

This PR closes #374.

## Developer Testing

Verified that enabling 'Show Slope' allows correctly displays.
Checked that enabling 'Show Slope' can only be done when less than 3 series are enabled.
Ensured slope calculations only occur when 'Show Slope' is enabled.
Checked that it automatically disables 'Show Slope' when 3 or more series are enabled.

## Reviewer Testing
Observe correct slope calculations when toggling 'Show Slope' on and off.
Ensure there are no unexpected UI issues or crashes when interacting with the parameter panel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/378)

## Summary by CodeRabbit

- **New Features**
	- Added an option to display the slope of a linear approximation for plotted data series in dashboard items.
	- Introduced controls to enable slope display for both plot and standard display widgets.

- **Improvements**
	- Slope display controls are dynamically enabled or disabled based on the number of selected series.
	- Numerical readout labels now center-align and append slope values when enabled for clearer data presentation.


ChrisYx511:
@ChrisWu0000 Will review once tests pass, please review linter errors.

ChrisYx511:
The linter problems seem to have to do with the prototype. I am suggesting deleting the prototype directory from the source tree.
Thoughts? @zangjiucheng

zangjiucheng:
@ChrisYx511 I am try to figure it out before merge it. I will delete from lint dir, but I don't know why I cannot reproduce same error on local. Flake8 will successful pass it on local.

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -53,8 +53,16 @@ def add_parameters(self):
         series_param = SeriesChecklistParameter()
         limit_param = {'name': 'limit', 'type': 'float', 'value': 0}
         offset_param = {'name': 'offset', 'type': 'float', 'value': 0}
-        return [series_param, limit_param, offset_param]
-
+        show_slope_param = {'name': 'Show Slope', 'type': 'bool', 'value': False}
```
ChrisYx511, line 56: Specify that this is change in y per second (dt = 1s)

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -162,6 +175,13 @@ def on_data_update(self, stream, payload):
                            if self.points[item] else 0 for item in self.series]
             for v in last_values:
                 current_values += f"[{v: < 4.4f}] "
+        if self.parameters.param('Show Slope').value():
+            slope_values = []
+            for stream in self.series:
+                slope = self.calculate_slope(self.times[stream], self.points[stream],
+                                            self.parameters.param('Number of Points').value())
+                slope_values.append(f"[{slope:.2f}]")
+            current_values += f"    Slope: {' '.join(slope_values)}"
```
ChrisYx511, line 184: Specify that this is per second (and confirm that it is actually per second)
```suggestion
            current_values += f"    Slope (/sec): {' '.join(slope_values)}"
```

ChrisYx511 (changes requested):
Core logic is good, please look at the implementation details below. If you disagree with any of the comments please do share your thoughts!
Thanks!

ChrisWu0000:
Most of the changes should be fixed, I'm not sure how to verify that the data is in seconds though

ChrisWu0000:
I'm a little confused what you mean by type annotation. Also the number of points should only appear once show slope is enabled, did you want me to do more than that?

ChrisYx511:
> I'm a little confused what you mean by type annotation. Also the number of points should only appear once show slope is enabled, did you want me to do more than that?

@ChrisWu0000 Oh I didn't see that it only shows when it's ticked. That works good! I would also include the check on __init__ though as when a dashboard savefile is loaded, the slope option might be ticked initially but the "Number of Points" option won't show until you untick and retick it. See image below:
image

Please see below for what I mean by title changes and type annotations.

ChrisWu0000:
Should be good now, let me know if I missed anything

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -53,7 +55,20 @@ def add_parameters(self):
...
+    
+    def _calculate_slope(self, times: list[float], points: list[float], num_points: int) -> float:
+        if len(times) < num_points or len(points) < num_points:
+            return np.nan
+        x = np.array(times[-num_points:])
+        y = np.array(points[-num_points:])
+        slope, _ = np.polyfit(x, y, 1)
+        return slope
+    
+    def on_show_slope_toggle(self, _, value):
```
ChrisYx511, line 70: Nit: Type annotations here

ChrisYx511 (changes requested):
Looks Good! Just some final nits and we are ready to merge!! :tada:
Sorry for holding it up for a bit exams have been rough LMAO.

Thanks for the work!

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -53,7 +55,20 @@ def add_parameters(self):
...
-        return [series_param, limit_param, offset_param]
+        show_slope_param = {'name': 'Show Slope of Linear Approx.', 'type': 'bool', 'value': False}
+        num_points_param = {'name': 'Slope: Num. of Points in Approx.', 'type': 'int', 'value': 10, 'limits': (2, None), 'visible': False} #How many points to use to calculate slope
+        return [series_param, limit_param, offset_param, show_slope_param, num_points_param]
+    
+    def _calculate_slope(self, times: list[float], points: list[float], num_points: int) -> float:
+        if len(times) < num_points or len(points) < num_points:
+            return np.nan
+        x = np.array(times[-num_points:])
+        y = np.array(points[-num_points:])
```
ChrisYx511, line 66: Nit: Data types for the numpy arrays
```suggestion
        x = np.array(times[-num_points:], dtype=np.float64)
        y = np.array(points[-num_points:], dtype=np.float64)
```

ChrisYx511: Or you can alternatively eliminate them altogether potentially by just going np.polyfit(times[-num_points:], points[-num_points:], deg=1)

ChrisYx511: Watch out for the duplicate i don't just mean replace the y LOL.

ChrisWu0000:
All good lol, hopefully ur exams went well. I added the annotations you suggested, I don't usually put annotations in my code so I might've missed some more. Let me know if you find any more.

sinks/dashboard/items/plot_dash_item.py
ChrisYx511: @ChrisWu0000 Missed this one!

ChrisYx511:
@ChrisWu0000 There's one more unresolved conversation in the previous pass and after that you should be good I'll double check.

ChrisYx511:
@ChrisWu0000 Could you also copy the code you wrote here into the Standard Display Item widget too? It works basically exactly the same the only thing you'd have to change is where the text is displayed.

ChrisYx511:
@ChrisWu0000 Hi! I'm back from the dead and realized I lowk made a little mistake defining requirements for this PR. 

I realized that setting a "Number of Points" doesn't mean anything for the us as the user has no idea what the data rate of the graph actually is. I think it would be much better UX to actually pick the number of seconds that the line averages over, where the number of seconds can be any value between (0, GRAPH_DURATION] (see config). 

Then, by default, it averages over the entire domain of the graph and the time is set to GRAPH_DURATION. 

I was also debating whether we should even let the user set over how many seconds do we take the linear approximation as quite honestly we only use the slope when there's something that's fairly slow (tank heating, a leak, etc.). The purpose of a slope is usually to calculate how much TIME a process is going to take. Logically, if it's something fast you wouldn't be able to see the slope in time to write it down and calculate a delta T during an operation anyways. And if it's something slow, then you can safely assume that the graph will represent a single system event the whole time (a leak, or tank heating) as GRAPH_DURATION is typically 30 seconds. Therefore, I might even be in favour of deleting this requirement completely of letting the user set the number of points / time. **Simplicity is key.** It would also make the code logic simpler.

@ChrisWu0000 I would like your thoughts. TL;DR I'm thinking either removing the number of points option entirely or turning it into a number of seconds. Leaning towards former due to better UX. Number of points doesn't tell the user anything. Number of seconds is OK but superfluous IMO.

CC: @zangjiucheng & DAQ People

ChrisWu0000:
I feel like deleting it is fine, as long as somewhere it says that it takes the entire duration or 30 seconds. Doesn't have to be on the UI maybe just mention it in the manual or something.

ChrisYx511:
Sounds good, I'll give you another OK to go ahead with this one I get approval in #electrical on Slack. I'll keep you updated.

ChrisYx511:
@coderabbitai review

ChrisYx511:
@ChrisWu0000 Please also have a look at the AI nits, especially the out-of-scope one that says that your standard display item code doesn't display the slope. 
image (https://github.com/waterloo-rocketry/omnibus/pull/378#pullrequestreview-2826943866)
Also we should consider using Polynomial.fit instead of np.polyfit because it's more numerically stable. That's what the official numpy docs say.
image
https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.fit.html#numpy.polynomial.polynomial.Polynomial.fit

ChrisYx511:
@ChrisWu0000 OK to remove the  number of points and just use the whole domain. Switch to Polynomial.fit.

ChrisYx511:
@ChrisWu0000 I think you are still not showing the slope in the standard display item LOL

ChrisWu0000:
Sorry my bad, I updated it now. The text value isn't centered, it wasn't before so I didn't bother to change it

sinks/dashboard/items/standard_display_item.py
```diff
@@ -89,12 +92,31 @@ def add_parameters(self):
...
+    
     def on_series_change(self, param, value):
+        if len(self.series) > 2:
+            self.parameters.param('Show Slope of Linear Approx.').setValue(False)
+            self.parameters.param('Show Slope of Linear Approx.').setOpts(enabled=False)
+        else:
+            self.parameters.param('Show Slope of Linear Approx.').setOpts(enabled=True)
+        if len(value) > 6:
+            self.parameters.param('series').setValue(value[:6])
         self.series = [value]
```
ChrisYx511, line 120: Why do you do this? This is the length of the series name as a Standard Display Item can only take one item. This ends up creating weird bogus series and is just buggy. Please test this. Please test all behaviour in Standard Display Item too.

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -53,9 +54,26 @@ def add_parameters(self):
...
 
     def on_series_change(self, param, value):
+        if len(value) > 2:
+            self.parameters.param('Show Slope of Linear Approx.').setValue(False)
+            self.parameters.param('Show Slope of Linear Approx.').setOpts(enabled=False)
+        else:
+            self.parameters.param('Show Slope of Linear Approx.').setOpts(enabled=True)
         if len(value) > 6:
             self.parameters.param('series').setValue(value[:6])
         self.series = self.parameters.param('series').childrenValue()
```
ChrisYx511, line 79: Similarly, why is this necessary?

sinks/dashboard/items/standard_display_item.py
```diff
@@ -216,7 +240,16 @@ def on_data_update(self, stream, payload):
                             t + config.GRAPH_STEP, padding=0)
         # For the numerical readout label
         self.data = float(point)
-        self.numRead.setText(f"{self.data:.{self.decimals}f}")
+        if self.parameters.param('Show Slope of Linear Approx.').value():
+            slope_values: list[str] = []
+            for stream in self.series:
+                slope = self._calculate_slope(self.times[stream], self.points[stream])
+                slope_values.append(f"{slope:.2f}" if not np.isnan(slope) else "N/A")
```
ChrisYx511, line 247: Why is this N/A where as in the plot item it is `--`?

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -53,9 +54,26 @@ def add_parameters(self):
...
+    def _calculate_slope(self, times: list[float], points: list[float]) -> float:
+        if len(times) < 2 or len(points) < 2:
+            return np.nan
+        x = np.array(times, dtype=np.float64)
+        y = np.array(points, dtype=np.float64)
+        # Use Polynomial.fit to fit a linear polynomial (degree=1)
+        p = np.polynomial.Polynomial.fit(x, y, deg=1)
+        # The slope is the coefficient of the linear term
+        return p.convert().coef[1]
+
```
ChrisYx511, line 69: good!

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -162,6 +180,12 @@ def on_data_update(self, stream, payload):
                            if self.points[item] else 0 for item in self.series]
             for v in last_values:
                 current_values += f"[{v: < 4.4f}] "
+        if self.parameters.param('Show Slope of Linear Approx.').value():
+            slope_values: list[str] = []
+            for stream in self.series:
+                slope = self._calculate_slope(self.times[stream], self.points[stream])
+                slope_values.append(f"[{slope:.2f}]" if not np.isnan(slope) else "--")
+            current_values += f"    Slope (/sec): {' '.join(slope_values)} "
```
ChrisYx511, line 189: Shouldn't this be under the if len(...)?

ChrisYx511 (changes requested):
Overall, we're in the home stretch! Please be careful when you're testing your changes however. This should be the final review once this is done I'll have one last look at it and we can merge!

sinks/dashboard/items/plot_dash_item.py
ChrisWu0000: I don't understand why this would need to be in an if len()? its been fine in all my testing

ChrisYx511: Oh whoops I did not realized that was gated by param value my bad you're good. Thanks for pointing it out!

sinks/dashboard/items/standard_display_item.py
```diff
@@ -216,7 +233,16 @@ def on_data_update(self, stream, payload):
                             t + config.GRAPH_STEP, padding=0)
         # For the numerical readout label
         self.data = float(point)
-        self.numRead.setText(f"{self.data:.{self.decimals}f}")
+        if self.parameters.param('Show Slope of Linear Approx.').value():
+            slope_values: list[str] = []
+            for stream in self.series:
+                slope = self._calculate_slope(self.times[stream], self.points[stream])
+                slope_values.append(f"{slope:.2f}" if not np.isnan(slope) else "--")
```
ChrisYx511, line 240: nit: Let's make NAN "[ -- ]" for consistency.

sinks/dashboard/items/plot_dash_item.py
```diff
@@ -162,6 +180,12 @@ def on_data_update(self, stream, payload):
                            if self.points[item] else 0 for item in self.series]
             for v in last_values:
                 current_values += f"[{v: < 4.4f}] "
+        if self.parameters.param('Show Slope of Linear Approx.').value():
+            slope_values: list[str] = []
+            for stream in self.series:
+                slope = self._calculate_slope(self.times[stream], self.points[stream])
+                slope_values.append(f"[{slope:.2f}]" if not np.isnan(slope) else "--")
```
ChrisYx511, line 187: nit: Let's make NAN "[ -- ]" for consistency.

ChrisYx511 (approved):
LGTM! Fix any nits before merging. Would ideally require unit tests in the future but we have yet to figure out the best way of doing that for dashboard so tbd.

Thank you for your work and sorry it took so long!

ChrisWu0000:
Changed it. No worries about taking a long time, I know ur pretty busy

Issue #394, opened 2025-06-17
Add support for CAN 2.0B Fields to `field_definitions.py` to standard data_processing script


ChrisYx511:
ETA: End of June

Problem: Currently, the data processing script, v1, does not support processing CAN 2.0 B messages because it does not know the new message headers.
Solution: Modify `field_definitions.py` such that a CSV file can be generated with the values from the new CAN subsystem. For reference on CAN 2.0B messages, see attached docs && reach out to @JasonBrave if necessary.

For log format (Login with UWaterloo Sharepoint): https://uofwaterloo.sharepoint.com/:t:/s/tm-eng-waterloorocketry/EST5ha0n_bdKhFuX00Dn3FcBD7rKxWMjec-Kfa32LfHbfQ?e=4JbIdY

For message types: https://docs.waterloorocketry.com/avionics/rocketcan/index.html

PR #397, opened 2025-06-27
refactor: update CAN field definitions and processing logic

zangjiucheng:
## Description

To support the New CAN Message, add `general_fields`, which is used in the shared `msg_type` in different boards, for example, for `GENERAL_BOARD_STATUS`. 

Also, update the CAN field definition, align with [packet format](https://docs.waterloorocketry.com/avionics/rocketcan/packet-format.html)

Example use: 
```py
# tools/data_processing/can_field_definitions.py

general_fields = [
    {
        "temp_name": "${board_id}_board_status",
        "signature": {'msg_type': 'GENERAL_BOARD_STATUS'},
        "fields": ["data.general_error_bitfield", "data.board_error_bitfield"]
    }
]
```

Change Log: 
1. Update `match()` in `CanProcessingField` class to return a wrapped CanProcessingField object instead of just boolean. 
2. Initial the `general_fields` structure. 

This PR is contribute to #394.

## Developer/Reviewer Testing

Try to parse with new log with CAN 2.0B to see if there is something break.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/397)

## Summary by CodeRabbit

* **New Features**
  * Enhanced support for dynamic CSV field naming based on board type, allowing more flexible data processing.
  * Added new CAN field definitions and updated existing ones to use a new naming convention.

* **Bug Fixes**
  * Improved matching logic for CAN fields, ensuring that unmatched fields are handled more accurately.

* **Refactor**
  * Updated internal logic to use new matching patterns and field definitions for improved maintainability.


tools/data_processing/can_processing.py
```diff
@@ -40,9 +40,12 @@ def get_can_lines(infile: IO, cols=[], msg_packed_filtering="behind_stream") ->
             # we check if the payload matches any of the fields we're tracking, and if it does, we update the current_info dictionary
             matched = False
             # WARNING: we need to loop through all the fields, becuase columns can have the same signatures and matching patterns, but we're taking differnt data out of them so we need to check multiple times
+            field:CanProcessingField
```
ChrisYx511, line 43: is this the nicest way to define this type annotation? not doubting, just genuinely curious :D

tools/data_processing/can_field_definitions.py
```diff
@@ -26,20 +29,26 @@ def __repr__(self):
...
                 nested_key, running_key = running_key.split(".", 1)
                 if nested_key not in checking:
-                    return False
+                    return None
                 checking = checking[nested_key]
             if running_key not in checking or checking[running_key] != value:
-                return False
-        return True
+                return None
+        return matched_instance if 'matched_instance' in locals() else self
```
ChrisYx511, line 51: I don't really understand the use of `locals()` here. Why can't we just extract matched instance as a nullable? Also what was wrong with previously returning boolean? Why do we have to refactor this? 😭

zangjiucheng: 1. The reason why, instead of boolean, we use an object is because we have a new `general_fields` which is an extension of `auto_fields`. For example, `general_error_bitfield` could be in any board; we don't want to add it one by one. In such a case, you will need a distinct `csv_name`, which will be extracted in the match function. Boolean won't export enough info to us when we go back to the extract function loop. 
So I design it match will return a wrapped matched object (If None, then not matched)

2. You can't return None here, because None has another meaning here. match_instance is only active to wrap a matched object when it detects it from `general_field` (I use the template string to achieve that). 
The reason why it will need to check if in locals() is because of efficiency. Here, I only need to do the `deepcopy` wrap when it's a general board information. You can wrap up every time, but we don't have to. I believe locals() in later than 3.11 is mostly O(1)? Not sure? But it's much less costly than doing a deepcopy of memory manipulation every time.

tools/data_processing/can_processing.py
ChrisYx511: Nope, it's Ok then. Missing space after `:`?

tools/data_processing/can_field_definitions.py
ChrisYx511: Ok but why can't we do this instead? Just curious 🧐
I'm probably missing something.

```python
matched_instance: Self | None = None
for key, value in self.matching_pattern.items():
    """ Rest of logic """
    # You would return None somewhere in here
    # .....
return matched_instance if matched_instance else self
```

zangjiucheng: Nope? I don't think so? (unless we do `matched_instance = None` first? I don't like to initial it here, it kinda breaks what `matched_instance` convenience here from my spaghetti idea lol)

```sh
UnboundLocalError: cannot access local variable 'matched_instance' where it is not associated with a value
```

ChrisYx511: I still think it's important to move that code out. I'll unblock you but unless you feel super strongly about `local()` I would really move it out.

ChrisYx511 (approved):
I'm going to approve but I would still fix the `local()` point mentioned.

tools/data_processing/can_processing.py
zangjiucheng: Sorry, I forgot this. I will add it in the next PR.

PR #400, opened 2025-07-09
Add Flowchart to explain Omnibus system overall

zangjiucheng:
As TITLE


zangjiucheng:
Thanks @Joe-Joe-Joe-Joe for pointing out the confusing part. I crafted this based on my experience, so it doesn't have an overall structure. All you mention is useful, I will fix that asap. 

Here is the link to the file: https://drive.google.com/file/d/1xVqHo8uYOuc05B5DGjRMrLu2a0MYKW2g/view?usp=drive_link

zangjiucheng:
> Don't call it the DAQ server if the server can be run from any node

Fixed to "Omnibus Server", yes, it can start from anywhere. 

> Don't understand the path from TIMESTAMP.log to Other Sources?

Fixed, it will go to the replay_log used to replay the message on ZMQ. 

>Why do Sources run through Parsley? Parsley is itself a source?
structure should be smth like Rocket bus->DAQ CAN support/logger board->Parsley->Sources

Yes, that sounds clearer.

>"RTOS" is misleading, most of omnibus runs on a regular Windows machine afaik

Fixed to "Data Stream"

> RLCS is missing entirely even though it is one the most important sources

I will try to draft something here, but maybe I need help from @Joe-Joe-Joe-Joe  and @JasonBrave to check on this. I don't have any experience with RLCS system, sorry :( 

> How many sinks are there? I feel like you could just list every sink. Same with sources

There are a total of 6, but `gpsd` is deprecated instead to use `interamap`, I have add the other two.
In the diagram, I really cannot show how each command works. I added a note on each of them, and based on that, for end-user will know which tool they need and look up the docs for more details.

> Does it use ZMQ between the sources and server as well? I assume so?

No, here is directly connected, so no ZMQ come in here.

### I will post a fixed diagram here to give more information. Also, I will add the docs link to the README. I think that will be useful.

zangjiucheng:
IMG_20231224_220015429_HDR

zangjiucheng:
omnibus-Page-1 drawio

PR #406, opened 2025-07-17
Allow string value comparisons on dynamic text

ChrisYx511:
## Description

I changed X and Y to accomplish Z.

This PR closes #XXX.

## Developer Testing

Here's what I did to test my changes:

- Ran A
- Ran B
- Ran C

## Reviewer Testing

Here's what you should do to quickly validate my changes:

- Run D and check output

PR #409, opened 2025-07-19
Parse new bitfields and fix dynamic text timeout in dashboard

ChrisYx511:
## Description

I increased the EXPIRED_TIME constant from 1.2 to 2.5 seconds and refactored the CAN parser error handling logic to use string comparison against "E_NOMINAL" instead of bitfield checks. This change simplifies error detection by treating error fields as enumerated strings rather than bitfields.
This PR depends on: https://github.com/waterloo-rocketry/omnibus/pull/406

## Developer Testing

- Tested dashboard items with the increased expiration time to ensure data persistence works correctly
- Verified CAN parser handles GENERAL_BOARD_STATUS messages with string-based error fields
- Confirmed error series generation works when general_bits and board_bits are not "E_NOMINAL"

## Reviewer Testing

- Send a GENERAL_BOARD_STATUS CAN message with error fields set to values other than "E_NOMINAL" and verify error series are generated
- Check that dashboard items now persist for 2.5 seconds before expiring

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/409)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for board status messages to more accurately display error states.
  * Extended the expiration timeout for dynamic text, so text remains visible longer before indicating expiration.


ChrisYx511:
@coderabbitai review

sinks/dashboard/parsers.py
```diff
@@ -130,20 +130,17 @@ def can_parser(payload):
...
+
     if message_type == "GENERAL_BOARD_STATUS":
-        # Pull the bitfields (default to 0 when missing)
-        general_bits = payload.get("general_error_bitfield", 0)
-        board_bits   = payload.get("board_error_bitfield",   0)
-    
-        # Exit early unless at least one bit is set
-        if general_bits or board_bits:
-            topic = f"{board_type_id}/{board_inst_id}/ERROR"
+        general_bits = data.get("general_error_bitfield")
```
zangjiucheng, line 136: Do you want to apply nit suggestion? Since the parser fails, won't crash the parse here with the default value. 

```diff
-   general_bits = data.get("general_error_bitfield")
-   board_bits   = data.get("board_error_bitfield")
+   general_bits = data.get("general_error_bitfield", "E_NOMINAL")
+   board_bits   = data.get("board_error_bitfield", "E_NOMINAL")
```

zangjiucheng (changes requested):
Pls review the nit comment here.

sinks/dashboard/parsers.py
ChrisYx511: I feel like we should not output E_NOMINAL when sometjing is not nominal 💭 ... I'll think about it...

zangjiucheng: Oh.... That also makes sense. It would throw an Error if they do some debugging on these two fields. I won't block on this, but I will have a review on canlib.

zangjiucheng (approved):

PR #417, opened 2025-07-31
fix: wait for rt_parser thread to stop before exiting application

vincentjguo:
Wait for rt_parser thread to exit before killing main thread

To test, start real time parsing and exit without stopping the parser. There should be no warnings of orphaned threads

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/417)

### DIFF
--- a/sinks/interamap/interamap.py
+++ b/sinks/interamap/interamap.py
@@ -67,6 +67,7 @@ def interamap_driver():
         if should_force_close:
             event.accept()  # Allow close without asking
         elif window.confirm_quit():
+            window.quit()
             event.accept()
         else:
             event.ignore()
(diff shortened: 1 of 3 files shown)


zangjiucheng:
Nice Fix! Yes it solving the thread issue. But I got following error, it's not a big deal since not error anymore but if it could be addressed would be even better! 

```
QThread::start: Thread termination error (No such process)
```

vincentjguo:
> Nice Fix! Yes it solving the thread issue. But I got following error, it's not a big deal since not error anymore but if it could be addressed would be even better!
> 
> ```
> QThread::start: Thread termination error (No such process)
> ```

I couldn't reproduce this but I'm pretty sure its because we terminated the thread when it already exited the event loop. Removed it, try now?

zangjiucheng (approved):
Yes that's the issue, LGTM. Nice fix.

PR #422, opened 2025-09-24
Added uv and removed setup.py

ChrisYx511:
## Description

This PR closes #XXX.

## Developer Testing

- Ran A
- Ran B
- Ran C

## Reviewer Testing

- Run D and check output


ChrisYx511:
This change is part of the following stack:

- #422 ◀
    - #423
        - #424

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

pyproject.toml
```diff
@@ -1,9 +1,65 @@
...
+    'Topic :: Scientific/Engineering',
+]
+
+authors = [
+    {name = "Waterloo Rocketry", email = "contact@waterloorocketry.com"}
+]
+
+requires-python = "~=3.11.0"
+
+dependencies = [
```
patrick-gu, line 39: Not that it was the case before but it would be good to put versions on these dependencies!

pyproject.toml
```diff
@@ -1,9 +1,65 @@
...
+    "autopep8",
```
patrick-gu, line 40: Could be added as a dev dependency

patrick-gu (approved):
I love `uv` and I think this will be a welcome addition!

pyproject.toml
ChrisYx511: This should be handled by the uv lockfile. I don't see much value imposing a minimum version considering I generated the uv lock like today AND it is generally bad practice to impose a max version except if there's something broken. Versions should be fixed regardless because of locking and will only be updated with a dedicated uv lock --upgrade-package command.

patrick-gu: I also commented there, but I think here the value of imposing a minimum version is to declare that we need features introduced in certain versions, and imposing a maximum version prevents breaking changes. It is [recommended](https://docs.astral.sh/uv/concepts/resolution/#lower-bounds) to declare a lower bound - I think the potential issue being if a direct dependency which relies on a upper bounded version of another direct dependency, which may give us a lower version than we had expected.

patrick-gu: Anyways, I don't think this is too consequential, so I think you should do as you see fit

pyproject.toml
ChrisYx511: Will do 🫡

PR #477, opened 2026-03-26
update ljm and ni message format

mathcslearner:
## Description

I updated message version from 2 to 3 in the LJM and NI sources.

This PR is related to task # 126 https://github.com/waterloo-rocketry/2025-software-issues/issues/126

## Developer Testing

Ran pytest

## Reviewer Testing

Run pytest

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/477)

## Summary by CodeRabbit

* **Bug Fixes**
  * Relative timestamps in DAQ messages are now seconds (float) instead of nanoseconds (int), improving timing semantics and consistency.
  * Period, reset, and timestamp generation now use seconds-based timing to ensure accurate inter-sample timing.
  * Message format version incremented to v3 to reflect the new timestamp convention.

* **Tests**
  * V3 tests updated to use the new relative_timestamps field.

### DIFF
--- a/src/sources/fakeni/main.py
+++ b/src/sources/fakeni/main.py
@@ -47,11 +47,11 @@ try:
                     relative_last_read_time_v3 + READ_PERIOD_v3 * i
                     for i in range(READ_BULK)
             ]
-            timestamps_key = "relative_timestamps_seconds"
+            timestamps_key = "relative_timestamps"
         else:
             relative_timestamps = list(range(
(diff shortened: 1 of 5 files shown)


Qubik65536 (changes requested):
It should be `relative_timestamps` instead of `relative_timestamps_seconds`. Otherwise we are good to go.

PR #479, opened 2026-04-02
fix(omnibus-ws): ensure message format is consistent with typescript implementation

ChrisYx511:
## Description

The WebSocket server's `handle_channel_message` handler previously accepted a single generic `data` argument and validated it was a 2-element list at runtime, which was fragile and inconsistent with how the TypeScript client sends messages (as a 2-element array unpacked into positional args).

This PR fixes the format by:
- Changing `handle_channel_message` to explicitly accept `timestamp` and `payload` as separate arguments, leveraging socket.io's argument unpacking
- Emitting payloads as tuples `(timestamp, payload)` instead of lists, matching the TS implementation
- Exporting `WS_ORIGINATED_SUFFIX` in `__all__` so consumers can reference it
- Fixing a pyright suppression comment (`reportImplicitRelativeImport` → `reportMissingImports`)

N/A

## Developer Testing

- Ran the WebSocket server locally and verified message relaying works end-to-end between ZMQ and WS clients
- Ran `pytest` — all tests pass

Tophat: https://waterloorocketry.slack.com/archives/C01535M46SC/p1775160205460759?thread_ts=1775151310.449209&cid=C01535M46SC

image
image
image
only missing branch but its exception throw

## Reviewer Testing

- Run `pytest`
- Start the server and a WS client; confirm messages flow correctly in both directions (ZMQ→WS and WS→ZMQ)

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/479)

## Summary by CodeRabbit

* **Refactor**
  * Socket.IO messages now use a positional timestamp + payload format (tuple), aligning client and server.
  * Client-side message handling and emissions updated to match the new positional shape.
  * Relay now forwards incoming WS messages consistently rather than conditionally dropping some.
  * Reduced informational relay logging to decrease console noise.
* **Chore**
  * Public exports extended to expose an additional suffix constant.

### DIFF
--- a/src/bridge/relay.py
+++ b/src/bridge/relay.py
@@ -58,16 +58,14 @@ def main(ws_url: str = "http://127.0.0.1:6767") -> None:
             print(f"[bridge] skipping message on '{msg.channel}' (originated from WS client)")
             continue
 
-        payload = [msg.timestamp, msg.payload]
+        payload = (msg.timestamp, msg.payload)
         try:
             sio.emit(msg.channel, payload) 
-            print(f"[bridge] relayed '{msg.channel}'")
         except (exceptions.ConnectionError, exceptions.BadNamespaceError) as e:
             print(f">>> Error sending message to WS server: {e}")
             reconnect(sio, ws_url)
             try:
                 sio.emit(msg.channel, payload) 
-                print(f"[bridge] relayed '{msg.channel}' after reconnecting")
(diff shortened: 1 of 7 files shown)


ShreyShingala (approved):
As long as the tests run properly this lgtm! 

Also #goat for that quick integration fix, the issue that we were supposed to send tuples got through is very unfortuante 😢

Qubik65536 (approved):

PR #492, opened 2026-05-18
YAML Configuration Refactor

john-jpet:
## Description
Refactored the LabJack LJM source to use a YAML-based configuration file instead of config.py. The new config.yaml format supports all existing constructs: stream settings (scans_per_read, scan_rate) and sensor definitions with linear and thermistor calibrations.

`main.py` now loads and parses `config.yaml` at startup using PyYAML, instantiating the same `calibration.py` classes as before. `config.py.example` is replaced by `config.yaml.example` with inline documentation for all supported fields.

Addresses: https://github.com/waterloo-rocketry/2025-software-issues/issues/141

## Developer Testing

- Ran `uv run python main.py` without a `config.yaml` present and confirmed the missing file error prints clearly to stderr
- Copied config.yaml.example to `config.yaml` and ran `uv run python main.py` - confirmed `Sensors: Test (V) on AIN0` printed correctly before failing on the LJM driver (not installed) and Omnibus server connection (not running), both expected without hardware
- Confirmed `config.yaml` does not appear in git status

## Reviewer Testing

- Copy `config.yaml.example` to `config.yaml` and run `uv run python main.py` from `src/sources/ljm` - confirm the sensor list prints correctly before failing on LJM/Omnibus
- Delete `config.yaml` and rerun, confirm the missing file error is clear and actionable
- Check git status after creating `config.yaml`, confirm it does not appear as untracked

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/492)

## Summary by CodeRabbit

* **Configuration**
  * LabJack data acquisition source configuration is now defined in YAML format instead of Python files
  * Configuration includes sensor definitions with linear and thermistor calibration types, connection modes, and stream parameters
  
* **Documentation**
  * Example YAML configuration template added to guide sensor setup and calibration options

[Review Change Stack](https://app.coderabbit.ai/change-stack/waterloo-rocketry/omnibus/pull/492?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/sources/ljm/.gitignore
+++ b/src/sources/ljm/.gitignore
@@ -1,1 +1,2 @@
 /config.py
+/config.yaml
--- a/src/sources/ljm/config.yaml.example
+++ b/src/sources/ljm/config.yaml.example
@@ -1,0 +1,53 @@
+#### EXAMPLE CONFIG ####
+## This is the documented sensor calibrations, as well as the port configurations in use on the DAQ Box.
+## To use the LabJack Source, duplicate this file and name it `config.yaml`, uncomment and modify as needed.
+## NOTE: If `config.yaml` is updated, please also update this file as required
+## Last modified 2026-05-18 - John Petruzzella (@jpetruzz)
+
+# LabJack Stream Configuration
+# Configured at 25 readings per read, and 1000 readings per second per channel.
+stream:
+  scans_per_read: 25
+  scan_rate: 1000
+
+# Sensor definitions
+# Each sensor requires:
+#   name:        Human-readable label (must be unique)
+#   channel:     LabJack AIN channel (e.g. AIN0)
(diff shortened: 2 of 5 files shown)


ShreyShingala (approved):
I don't see any big issues, so lgtm, tho ask @Qubik65536 to take a look too as he works with LabJack pretty often

ChrisYx511 (changes requested):
few things:
- we should break the yaml parsing code out of main ideally, and make it a nicer API that takes in a YAML string buffer and returns a list of Calibrations. IMO I think dumping it all into main is a little hard to navigate but we can debate this.
- the yaml parsing code should have unit tests. all new code should be tested.
- we should not break compatibility with `config.py` if both config.py and config.yaml exist, config.py should be prioritized
- please make sure to also include `types-pyyaml` to ensure that the YAML is strictly typed and checked

PR #499, opened 2026-06-11
feat: update ljm source with v3 DAQ messages for omnibus-ts compat

ChrisYx511:
## Description

This pull request updates the data format for DAQ messages to improve timestamp handling and ensure compatibility with new tools. The main change is switching from integer nanosecond timestamps to floating-point second timestamps, which simplifies downstream processing. The message format version is incremented to reflect this breaking change.

**Data format changes:**

* Changed the `DAQ_SEND_MESSAGE_TYPE` to use `relative_timestamps` (list of floats in seconds) instead of `relative_timestamps_nanoseconds` (list of ints in nanoseconds). Updated all related logic to generate and use these timestamps. [[1]](diffhunk://#diff-b7d2d16a9dea1514fca115c2dccad8213aa337a79a97f00fda5dc73e4d392d79L61-R66) [[2]](diffhunk://#diff-b7d2d16a9dea1514fca115c2dccad8213aa337a79a97f00fda5dc73e4d392d79L127-R140)
* Updated the calculation of `relative_last_read_time` to use integer nanoseconds for internal tracking, but convert to seconds for the new `relative_timestamps` field. [[1]](diffhunk://#diff-b7d2d16a9dea1514fca115c2dccad8213aa337a79a97f00fda5dc73e4d392d79L88-R88) [[2]](diffhunk://#diff-b7d2d16a9dea1514fca115c2dccad8213aa337a79a97f00fda5dc73e4d392d79L148-R150)

**Versioning:**

* Incremented `MESSAGE_FORMAT_VERSION` from 2 to 3 to indicate the breaking change in timestamp representation.

N/A

Will test today before CF.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/499)

### DIFF
--- a/src/sources/ljm/main.py
+++ b/src/sources/ljm/main.py
@@ -40,7 +40,7 @@ sender = Sender()
 CHANNEL = "DAQ"
 # Increment whenever data format change, so that new incompatible tools don't
 # attempt to read old logs / messages.
-MESSAGE_FORMAT_VERSION = 2  # Backwards compatible with original version.
+MESSAGE_FORMAT_VERSION = 3  # Backwards compatible with original version.
 
 
 class DAQ_SEND_MESSAGE_TYPE(TypedDict):
@@ -58,12 +58,12 @@ class DAQ_SEND_MESSAGE_TYPE(TypedDict):
     # }
     # 1.3 and 2.3 are the readings for each sensor at t0, 2.3 and 4.5 for t1, etc.
 
-    relative_timestamps_nanoseconds: list[int]
+    relative_timestamps: list[float]
     """
     Corresponding timestamps for each reading of every sensors, calculated from sample rate (dt_ns = 1/sample_rate * 10^9).
     There can be variation of +- 1ns for every point.
     Timestamps are based on initial time t_0 = time.time_ns(), meaning they should be always unique.
-    Unit is nanoseconds.
+    Unit is seconds.
     """
     # Example: [19, 22, 25] <- 1.3 and 2.3 from above was read at t0 = 19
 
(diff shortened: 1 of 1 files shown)


src/sources/ljm/main.py
```diff
@@ -58,12 +58,12 @@ class DAQ_SEND_MESSAGE_TYPE(TypedDict):
     # }
     # 1.3 and 2.3 are the readings for each sensor at t0, 2.3 and 4.5 for t1, etc.
 
-    relative_timestamps_nanoseconds: list[int]
+    relative_timestamps: list[float]
```
JasonBrave, line 61: have you considered floating point math causing weird issues, like 4.999999999 seconds?

ChrisYx511: tbh this is so much less of an issue than trying to serialize bigints over msgpack and dealing with them in TypeScript

the big issue is that JavaScript doesn't speak ints. It only has floats. and this level of imprecision is far less than our hardware imprecision tbh

ChrisYx511: like i highly doubt this is gonna be an issue since when we plot the data we're gonna convert to seconds anyways. This is just to index the data.

ShreyShingala (approved):
lgtm

Qubik65536 (approved):
I have no problem of the current implementation of timestamps (using float), so resolving it in anyway should be good.

src/sources/ljm/main.py
JasonBrave: ok

PR #517, opened 2026-07-05
Guard dashboard items against non-numeric (string/enum) streams

ShreyShingala:
## Problem
The dashboard crashes when a string/enum stream reaches a widget that assumes numeric data:

```
TypeError: can only concatenate str (not "int") to str   # PlotDashItem
ValueError: could not convert string to float: 'E_NOMINAL' # GaugeItem / ProgressBarItem
```

Same change was made to other files that had similar issues
- `plot_dash_item.py` — `point += self.offset`
- `gauge_item.py` — `float(point)`
- `progress_bar.py` — `float(point)`

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/517)

## Summary by CodeRabbit

* **New Features**
  * CSV/log outputs now include message metadata fields, improving traceability of parsed entries.
  * CAN data matching and field selection now support metadata-based identifiers for more accurate mapping.

* **Bug Fixes**
  * Dashboard widgets now ignore non-numeric incoming values instead of failing on string-like payloads.
  * CAN field peeking and parsing now use updated message signatures for more reliable output headers and matching.

### DIFF
--- a/src/dashboard/items/gauge_item.py
+++ b/src/dashboard/items/gauge_item.py
@@ -8,6 +8,7 @@ from .dashboard_item import DashboardItem
 from .registry import Register
 
 from decimal import Decimal
+import numbers
 
 @Register
 class GaugeItem(DashboardItem):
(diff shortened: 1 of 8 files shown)


ShreyShingala:
@coderabbitai review

Qubik65536 (approved):

PR #520, opened 2026-07-11
Pad decimal-minutes to 4 digits

ShreyShingala:
## Summary
Used to be zero-padded to only 2 digits but should be 4 digits to be represented correctly.

## Why 4 digits is correct (not 2)
`dmins` is a 16-bit field, so its width alone doesn't dictate a digit count, this is confirmed by how the two other places that consume it already treat it:

- **parsley** decodes it as a plain ten-thousandths integer (0-9999) — see the round-trip test at [`tests/test_message_definitions.py#L154-L161`](https://github.com/waterloo-rocketry/parsley/blob/34f3f92091c3a9a16946b7534111430ee048ee7d/tests/test_message_definitions.py#L154-L161) (`dmins == 1234` for a specific wire value) and the field definition at [`src/parsley/message_definitions.py#L32-L33`](https://github.com/waterloo-rocketry/parsley/blob/34f3f92091c3a9a16946b7534111430ee048ee7d/src/parsley/message_definitions.py#L32-L33).
- **omnibus/interamap** already computes real coordinates from it using `dmins / 600000` (i.e. `(dmins / 10000) / 60`) — see [`src/sinks/interamap/src/real_time_parser.py#L47-L51`](https://github.com/waterloo-rocketry/omnibus/blob/main/src/sinks/interamap/src/real_time_parser.py#L47-L51). That division only produces the correct minute fraction if `dmins` is treated as a 4-digit (ten-thousandths) value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/520)


Qubik65536 (approved):
LGTM

ChrisYx511 (commented):
Not blocking this but do have a question: Is this correctly implemented in Interamap? Can we ask an elec GPS board person just to make sure?

ShreyShingala:
> * /

Will confirm today

ShreyShingala:
https://github.com/waterloo-rocketry/docs/blob/main/software/omnibus/gpsd.rst

its deprecated

PR #521, opened 2026-07-11
GPS points with the own board

ShreyShingala:
Board_id was getting pulled from whatever CAN message last came off the receiver instead of the board that actually
finished sending its gps data. If board A's bundle completes on a tick where board B's message just came in, the point gets tagged as board B. Fixed by passing the board id in directly instead of grabbing it off whatever message happened to be last.

## Reviewer Testing

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/521)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPS data processing to consistently associate location updates with the correct board identifier.
  * Prevented GPS metadata from being incorrectly derived from unrelated message contents.

### DIFF
--- a/src/sinks/interamap/src/real_time_parser.py
+++ b/src/sinks/interamap/src/real_time_parser.py
@@ -16,7 +16,7 @@ except ImportError:
     from config import BoardID, BOARD_FIELDS # type: ignore
 
 
-def parse_gps_data(gps_data, data):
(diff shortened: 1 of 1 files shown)


Qubik65536 (approved):
LGTM

PR #522, opened 2026-07-11
Guard thermistor calibrate() divide-by-zero

ShreyShingala:
Self explanitory, guards a divide by 0

## Reviewer Testing

- Run D and check output

- - -
This change is [Reviewable](https://reviewable.io/reviews/waterloo-rocketry/omnibus/522)


Qubik65536 (commented):
Looks reasonable but regarding the comment: "disconnected", shouldn't be those just removed from config? Do we want to add something else that tells what's happening instead of just failing silently?

ChrisYx511 (approved):
this is great! thanks!

